PR #3, opened 2018-10-16
Docs for program stage in program rule

Birkbjo:


src/commonmark/en/content/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -1430,6 +1430,11 @@ objects:
 
       - **Program**
 
+      - **Trigger rule only for program stage**
+
+        If a program stage is selected, the program rule will only run for the selected program stage,
+        as opposed to being run for every program stage in the program.
```
Philip-Larsen-Donnelly: I think this sentence is OK, but it could be broken up to make it even easier to understand (without having to read it slowly a couple of times). Consider, e.g.:
If a program stage is selected in this field, the program rule will only run for that selected program stage. If this field is left empty, the program rule will apply to every program stage in the program.

Philip-Larsen-Donnelly (dismissed):
I'm being really picky here!

PR #12, opened 2018-10-30
DHIS2-975-option-group

Birkbjo:
User docs for maintenance option group and option group set

### DIFF
--- a/src/commonmark/en/content/user/configure-metadata.md
+++ b/src/commonmark/en/content/user/configure-metadata.md
@@ -3964,6 +3964,7 @@ Options that are grouped can be hidden or shown together in tracker and event ca
 
 **Option group sets** allows you to categorise multiple **option groups** into a set.
 The main purpose of the option group set is to add more dimensionality to your captured data for analysis.
+
 > **Note**
 >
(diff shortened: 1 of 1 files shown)


src/commonmark/en/content/user/configure-metadata.md
```diff
@@ -3946,19 +3948,22 @@ This way you can create a subset of options in an option set. The main purpose o
...
-6. Click **save**.
+6. Click **Save**.
 
 ### Create or edit an option group set
 
 <!--DHIS2-SECTION-ID:create_option_group_set-->
 
 **Option group sets** allows you to categorise multiple **option groups** into a set.
+The main purpose of the option group set is to add more dimensionality to your captured data for analysis.
 > **Note**
```
Philip-Larsen-Donnelly: You need to make sure there is a blank line before this "> **Note**" so that the markdown will render properly
In general blank lines seem to be incredibly important in commonmark, in order to distinguish different block of text. (This is a bit of a weakness of the specification)

Philip-Larsen-Donnelly (changes requested):
Blank line needed before "Note" (aside block), but otherwise looks good.

Philip-Larsen-Donnelly (approved):

PR #16, opened 2018-11-07
import-export-app(data export): update docs

adeelshahid:


Philip-Larsen-Donnelly (commented):
Hi Adeel. I think this update could be approved, but the text in the screenshot says "include descendant ...", should it not use the plural "include descendants ...". I know it sounds a bit picky, but I think this could cause unnecessary confusion if the singular is used.

Philip-Larsen-Donnelly (approved):

PR #35, opened 2018-11-28
Update data quality chapter.

jason-p-pickering:


Philip-Larsen-Donnelly (approved):
The changes all look fine to me, so this PR is approved.
 
Just one additional note: there are several "Tips" throughout the chapter that refer to clicking a "star icon" to do this or that. These tips should be reviewed as they are no longer true for the new UI.

PR #58, opened 2018-12-17
Adds info about new SQL_VIEW_MAX_LIMIT dropdown in settings-app

HendrikThePendric:
Added a new table row with info about the new SQL_VIEW_MAX_LIMIT dropdown in the settings-app. I've also updated odd/even class names in the subsequent table rows, but I don't actually know if that has any effect.

### DIFF
--- a/src/commonmark/en/content/user/system-settings.md
+++ b/src/commonmark/en/content/user/system-settings.md
@@ -29,54 +29,59 @@
 </blockquote></td>
 </tr>
 <tr class="even">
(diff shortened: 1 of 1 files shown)


Philip-Larsen-Donnelly (approved):

Philip-Larsen-Donnelly:
The class names don't really have any effect as we are not currently using them in the stylesheet for the documentation. I doesn't harm to keep them up to date for now, but I think in principle we don't need to explicitly set the class for each row - so we could eventually strip out all of these inline odd/evens without limiting our styling options.

PR #75, opened 2019-02-04
fix(settings-app): hide periods in analysis app

HendrikThePendric:
Added explanation for 4 new fields to the docs:
ANALYTICS_HIDE_DAILY_PERIODS
ANALYTICS_HIDE_WEEKLY_PERIODS
ANALYTICS_HIDE_MONTHLY_PERIODS
ANALYTICS_HIDE_BIMONTHLY_PERIODS

### DIFF
--- a/src/commonmark/en/content/user/system-settings.md
+++ b/src/commonmark/en/content/user/system-settings.md
@@ -55,22 +55,6 @@
 <td><p>Setting this value will determine which relative period is selected as the default in the analytics apps.</p></td>
 </tr>
 <tr class="odd">
-    <td><p><strong>Hide daily periods</strong></p></td>
-    <td><p>Hide daily periods in the analysis tools</p></td>
-</tr>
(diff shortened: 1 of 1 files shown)


Philip-Larsen-Donnelly (changes requested):
These fields are documented here under General Settings, but they are implemented in the Analytics Settings tab.

Philip-Larsen-Donnelly (approved):

HendrikThePendric:
Well spotted Phil,sorry. I simply searched for the string "Default relative period for analysis" because I knew these new fields should go below that field. I hadn't realised that "Default relative period for analysis" has two entries (both under General and un Analytics settings). Not sure if this is intentional either....

Philip-Larsen-Donnelly:
Thanks @HendrikThePendric. My guess it that the "Default relative period for analysis" should not be under General. I'll check and raise an issue if necessary.

PR #117, opened 2019-03-06
update 2.30 dashboard interpretations sort-order and rich-text support [DHIS2-3422]

joakimia:
Add docs for rich text support and sort order of interpretations for the 2.30 dashboard.


Philip-Larsen-Donnelly:
Isn't this the same change as the previous PR? #116

PR #132, opened 2019-03-20
DHIS2-5987 - Add information about data visualizer "open as map" feature

neeilya:


Philip-Larsen-Donnelly (changes requested):
The text and link look fine, but please separate each of the three lines here with spaces to avoid ambiguous interpretation of the markdown.

> Spaces are very important in markdown for helping to clarify where blocks begin and end. This _might_ render OK, but it will depend a bit on the markdown interpreter. It also may cause problems for our localisation platform.

Philip-Larsen-Donnelly (approved):

PR #135, opened 2019-04-01
Docs for event assignment filtering capability in apis and filter entities

ameenhere:
Added documentations for
DHIS2-6235 - Extend event apis to use user assignment for filtering
DHIS2-5445- EventFilters management
DHIS2-6056- Add event user assignment in TEIFilter
DHIS2-6241 - Extend tei apis to use user assignment on events for filtering


src/commonmark/en/content/developer/web-api.md
```diff
@@ -16464,6 +16476,236 @@ based on data element
...
+<td>"ouMode": "SELECTED"</td>
+</tr>
+<tr class="odd">
+<td>assignedUserMode</td>
+<td>To specify the assigned user selection mode. Possible values are CURRENT| PROVIDED | NONE | ANY </td>
+<td>"assignedUserMode": "NONE"</td>
+</tr>
+<tr class="even">
+<td>assignedUsers</td>
+<td>To specify a list of assigned users for events.</td>
```
Bekkalizer: How does the assigned users and the assigned user mode interact?

src/commonmark/en/content/developer/web-api.md
```diff
@@ -16464,6 +16476,236 @@ based on data element
...
+<td>lastUpdatedDate</td>
+<td>DatePeriod object containing a from and to date for filtering based on last updated date.</td>
+<td>{ &quot;from&quot;:&quot;2018-01-01&quot; , &quot;to&quot;: &quot;2019-01-01&quot;}</td>
+</tr>
+<tr class="even">
+<td>events</td>
+<td>To specify list of events</td>
+<td>"events"=["a3kGcGDCuk6"]</td>
+</tr>
+</tbody>
```
Bekkalizer, line 16744: We need eventDate and completedDate filters also

Bekkalizer: And as commented earlier, we need relative periods here. The trackedEntityInstance filters has a syntax that can be used as inspiration.

src/commonmark/en/content/developer/web-api.md
```diff
@@ -16464,6 +16476,236 @@ based on data element
...
+<td>To specify the assigned user selection mode. Possible values are CURRENT| PROVIDED | NONE | ANY </td>
+<td>"assignedUserMode": "NONE"</td>
+</tr>
+<tr class="even">
+<td>assignedUsers</td>
+<td>To specify a list of assigned users for events.</td>
+<td>"assignedUsers": ["a3kGcGDCuk7", "a3kGcGDCuk8"]</td>
+</tr>
+<tr class="odd">
+<td>trackedEntityInstance</td>
```
Bekkalizer: I dont think we have a usecase for this now. Perhaps we dont add it until we need it?

src/commonmark/en/content/developer/web-api.md
```diff
@@ -16464,6 +16476,236 @@ based on data element
...
+<td>"filters"=["a3kGcGDCuk6:LT:4"]</td>
+</tr>
+<tr class="even">
+<td>fields</td>
+<td>To specify fields that has to be part of the event listing response.</td>
+<td>"fields"="a3kGcGDCuk6,eventDate,dueDate"</td>
+</tr>
+<tr class="odd">
+<td>dataElements</td>
+<td>To specify dataElements that has to be part of the event grid query response.</td>
```
Bekkalizer: We also need a way of specifying the data value filtering. If you open the capture app you see how the searches will specify filterings for data values in data elements.

Bekkalizer (changes requested):

src/commonmark/en/content/developer/web-api.md
ameenhere: Shortest reply to all your comments is , I included the FULL capability of our APIs (normal + grid) in this model. Everything that is currently possible in our API as far as filtering/sorting/trimming/anything , its possible to save them in this model. The only thing I missed is the relative periods, which I am working on now. There were some parameters which were exclusively applicable for Grid API and vice versa. I have included both of them in such cases.

src/commonmark/en/content/developer/web-api.md
ameenhere: assignedUsers(if non empty) is considered when assignedUserMode=PROVIDED (or null). For all other assignedUserMode, assignedUsers should be empty (unspecified)

src/commonmark/en/content/developer/web-api.md
ameenhere: Yep. Can remove it. Included it just because we had the parameter in our event listing API

src/commonmark/en/content/developer/web-api.md
ameenhere: 1. isnt eventDate same as createdDate?
2. I can add completedDate in the filter definition. But the event listing APIs currently does not support filtering based on completedDate.

src/commonmark/en/content/developer/web-api.md
Bekkalizer: Ok sounds good - can you document the contradictions? Will the assignedUserMode overrride any provided users unless the mode is PROVIDED?

src/commonmark/en/content/developer/web-api.md
Bekkalizer: 1. Both a created date and event date exists for an event - The event date is the date we usually reference and use for filtering.
2. What dates are supported for this filtering at the moment?

src/commonmark/en/content/developer/web-api.md
Bekkalizer: Ok, so there is a syntax for querying data element values. Could you add an example here perhaps?

src/commonmark/en/content/developer/web-api.md
ameenhere: Will do!

src/commonmark/en/content/developer/web-api.md
ameenhere: In the api parameter, we have 3 set of dates. 1) startDate/endDate 2) dueStartDate/dueEndDate 3)lastUpdatedStartDate/lastUpdatedEndDate. these 3 can be used to filter the results in the API as of now

src/commonmark/en/content/developer/web-api.md
ameenhere: "dataElement=a3kGcGDCuk6:GT:6" this works for the query end point.

src/commonmark/en/content/developer/web-api.md
Bekkalizer: ok great - lets document how to use these for now.

Bekkalizer (approved):

PR #148, opened 2019-04-24
Externalize server URL property

zubaira:


bobjolliffe:
Good to see this config parameter.  Can I suggest in the documentation that you use a different example than http://localhost:8080?  Other than developer's laptops this is unlikely to be the form required in practice.  And using the localhost example doesn't make it clear that this is for the externally accessable url so folk might be confused.

Maybe just use https://play.dhis2.org/dev as an example?

Philip-Larsen-Donnelly (approved):

PR #158, opened 2019-04-29
New expression features for indicators & program indicators

jimgrace:


src/commonmark/en/content/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -1067,6 +1067,36 @@ functions. The functions can be applied to data elements and attributes:
 <td><p>(dataElement, condition)</p></td>
 <td><p>Useful only for enrollment program indicators. Counts the number of data values that matches the given condition criteria for the given program stage and data element in the course of the enrollment. The argument data element is supplied with the #{programStage.dataElement} syntax. The condition is supplied as a expression in single quotes, for example '&lt;10' if only the values less than 10 should be counted.</p></td>
 </tr>
+<tr class="odd">
```
Philip-Larsen-Donnelly, line 1070: All of the previous functions have a "d2" namespace. What is the distinction with the new functions? i.e. the reason for the apparent inconsistency?

Philip-Larsen-Donnelly (commented):
Looks good to me! One small comment added.

src/commonmark/en/content/user/configure-programs-in-the-maintenance-app.md
jimgrace: The new functions are common to all expression parsing including for indicators and program indicators (and in the future validation rules and predictors). The d2 functions are for program indicators and rules only. We are introducing the simpler syntax across the board in all expressions rather than propagating the d2 namespace across the board.

larshelge (approved):

PR #160, opened 2019-04-30
feat(beta-reporting-app): add docs for beta reporting-app

HendrikThePendric:
I've copied the chapter of the old reports-app and have updated text and screenshots where needed. I've tried to consistently refer to the new reports app as "Reports app (Beta)" since this is what it is going to be called in the menu.

I have been able to review the markdown document pretty well, but I have not built the HTML docs locally, so I couldn't verify that the chapters has been added to the tables of content correctly, and that the referable tags are working.


Philip-Larsen-Donnelly:
Thanks Hendrik, I'll take a look!

Philip-Larsen-Donnelly (approved):

PR #175, opened 2019-06-12
Typo fixes and update re: discontinued Java PPA

WStringer2:
* sample command to install postgis referenced dhis instead of dhis2
* The referenced Java PPA has been discontinued.  Added a note to use openjdk instead
* Update chown command to reference the appropriate folder
* Update chown command to use dhis instead of test user


Philip-Larsen-Donnelly (approved):
Looks good. Thanks!

PR #188, opened 2019-07-26
Fix incorrect parameter name

huzvak:
DHIS2-5557

### DIFF
--- a/src/commonmark/en/content/user/configure-programs-in-the-maintenance-app.md
+++ b/src/commonmark/en/content/user/configure-programs-in-the-maintenance-app.md
@@ -1233,12 +1233,12 @@ variables:
 <pre><code>V{program_stage_id} == &#39;YPSSfbmAtt1&#39;</code></pre></td>
 </tr>
 <tr>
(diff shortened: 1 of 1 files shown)


src/commonmark/en/content/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -1248,7 +1248,7 @@ A filter that uses the "Reporting period end" variable to only include
 women who has an LMP that would be in the first
```
Philip-Larsen-Donnelly, line 1248: Need to change the quoted variable in this sentence.

Philip-Larsen-Donnelly (changes requested):
small change needed.

src/commonmark/en/content/user/configure-programs-in-the-maintenance-app.md
huzvak: Fixed

Philip-Larsen-Donnelly (approved):

PR #197, opened 2019-08-07
docs(fieldmask): add fieldmask for data elements and tea

Birkbjo:
Docs for https://jira.dhis2.org/browse/DHIS2-4717

### DIFF
--- a/src/commonmark/en/content/user/configure-metadata.md
+++ b/src/commonmark/en/content/user/configure-metadata.md
@@ -622,16 +622,18 @@ objects:
     about how the data element is measured and what its purpose is.
 
 8. (Optional) In the **Field mask** field, you may type a template that's used to provide
-   hints for correct formating of the data element in Capture and Tracker Capture apps.
-   The following are special characters that can be used in the mask to match more than one character.
+   hints for correct formatting of the data element in Capture and Tracker Capture apps.
+   The following are special characters that can be used in the mask. The special characters match exactly one character of the given type.
 
    | Character     |    Match       |
    | ------------- |----------------|
-   |       \#      |     digit      |
-   |       x       |lower case letter|
-   |       X       | capital letter |
-   |       \*      | any character  |
+   |      \\d      |     digit      |
+   |      \\x      |lower case letter|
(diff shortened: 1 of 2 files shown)


src/commonmark/en/content/user/configure-metadata.md
```diff
@@ -621,6 +621,18 @@ objects:
     element. Be as precise as possible and include complete information
     about how the data element is measured and what its purpose is.
 
+8. (Optional) In the **Field mask** field, you may type a template that's used to provide
+   hints for correct formating of the data element in Capture and Tracker Capture apps.
+   The following are special characters that can be used in the mask to match more than one character.
```
Philip-Larsen-Donnelly: If they can be used to match "more than one character", shouldn't the match be "digit(s)", "lower case letter(s)", etc?

Also, to be precise I think you probably want to way "one or more characters" rather than "more than one character", no?

Philip-Larsen-Donnelly: (I might also be useful with a couple of examples)

Philip-Larsen-Donnelly (changes requested):

src/commonmark/en/content/user/configure-metadata.md
Birkbjo: I was actually quite unsure how to phrase it, as the special characters match exactly one character of the given type. Ie #### is exactly 4 digits. Examples will be added. I just need to find some good ones,  as we do not support variable length of the mask- so phone number (which is a natural one) is not really a good example.

Philip-Larsen-Donnelly: OK. then instead of:
> The following are special characters that can be used in the mask to match more than one character.

I would probably try:
> The following are special characters that can be used in the mask. The special characters match exactly one character of the given type.

Philip-Larsen-Donnelly: (sometimes things like this are clearer when you break it down into smaller sentences)

Philip-Larsen-Donnelly (approved):

PR #209, opened 2019-08-26
Added sub-chapter on configuring Tracker programs

karolinelien:
PR probably needs some work and formatting but includes subchapter on configuring Tracker programs. Currently not showing up in preview (not sure why) but submitting as PR so someone with markdown experience can review.


Philip-Larsen-Donnelly (approved):
I've only skimmed, but looks great. We can tweak if necessary later. Let's get this in!
Thanks Karoline!!

PR #212, opened 2019-09-01
docs: Update default merge mode on metadata import

SferaDev:
The default value is actually REPLACE and it could lead to confusion. You can test it through a POST call to the ```/metadata``` endpoint in the importParams response.

See: https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/MetadataImportParams.java#L106

Feel free to re-order the description column if you see fit!


Philip-Larsen-Donnelly (approved):
Thanks Alexis!

PR #218, opened 2019-09-13
feat: Bulk get approval status DHIS2-4413

jimgrace:


src/commonmark/en/content/developer/web-api.md
```diff
@@ -5969,6 +5969,90 @@ from the following
...
+              mayReadData: true
+          },
+          state: "UNAPPROVED_READY",
+          wf: "rIUL3hYOjJc"
+      }
+    ]
+
+The returned parameters are:
+
+<table>
```
Philip-Larsen-Donnelly: It would be great if we could use a GFM table here instead! 
(In general, I want people to use GFM (pipe-based) tables wherever possible as I think they are cleaner, and they really help with translation).
e.g.:

**Bulk data approval returned parameters**
| Return Parameter | Description |
| :- | :--- |
| aoc | Attribute option combination identifier |
| pe | Period identifier |
| ou | Organisation Unit identifier |
| permissions | The permissions: 'mayApprove', 'mayUnapprove', 'mayAccept', 'mayUnaccept', and 'mayReadData' (same definitions as for get single approval status.) |
| state | One of the data approval states (same as for get single approval status.) |
| wf | Data approval workflow identifier |

Philip-Larsen-Donnelly (changes requested):
Can you please try with the GFM table?

Otherwise, it looks fine. (Though I'm not really a fan of words like "Unapprove" and "Unaccept"! ;) )

src/commonmark/en/content/developer/web-api.md
jimgrace: I was just copying what was nearby. But if mixing table types doesn't matter in a section, I agree that we should be using markdown tables for new content (and maybe converting old HTML ones too.)

Philip-Larsen-Donnelly: Thanks Jim. I think mixing is fine! I think the more GFM tables we have, the better.

I hope to convert the old ones if/when time permits... (It can be automated a little bit, but not very easily, and usually needs some manual tweaking)

PR #224, opened 2019-09-19
docs: update import export app docs after refactoring

Mohammer5:
I've made some changes in this document:

1. I've changed the number in the lists I've worked on to only use `1.`. If that's not supported by our documentation-generating scripts, please let me know and I'll revert it
1. I've Introduced `Importing data` and `Exporting data` h2 headlines to group import and export pages
1. I've added the `Import progress logger` which describes how the logger can be opened/closed.
Previously the last item in the lists of importing data said that the logger will be opened, which is false.
1. I've switched the order of `Event import` and `GML import` to reflect the order in the import export app
1. I've extended the lists of editable fields on the `Metadata` and `Data import` pages
1. I've replaced the "Category Id Scheme" with "Id scheme" option on the `Data export` page
1. Add section references to all sections


Philip-Larsen-Donnelly:
1. The use of `1.` is supported :+1: 
2-6. Sounds good. I'll review!

src/commonmark/en/content/user/importexport-app.md
```diff
@@ -149,8 +175,25 @@ import.
...
+
+1.  Upload a file using the *GML* (Geographic Markup Language) format.
+
+2.  Choose **Dry Run** . *Yes* will do a test import without importing
+    any data into the database.
+
+1.  Click on the **Import** button which will upload the file and start the
+    importing process.
 
 ## Metadata Import Summaries
```
Philip-Larsen-Donnelly: Would it make sense for this now to be a h3 level?
Should it also be simply "Import Summaries"?
(maybe it also belongs next to the "Import process logger" section?)

Mohammer5: Oops, yes, that should be an h3 as well.
I'll move it below the "Import process logger" section as you've suggested

src/commonmark/en/content/user/importexport-app.md
```diff
@@ -27,8 +27,12 @@ which can be found below.
 
 ## Importing data
 
+<!--DHIS2-SECTION-ID:importing_data-->
+
 ### Import progress logger
 
+<!--DHIS2-SECTION-ID:import_process_logger-->
```
Philip-Larsen-Donnelly: progress or process?
(doesn't really matter, but nice to be consistent).

Mohammer5: Good catch

Philip-Larsen-Donnelly (approved):

src/commonmark/en/content/user/importexport-app.md
Mohammer5: Done

src/commonmark/en/content/user/importexport-app.md
Mohammer5: Fixed

PR #227, opened 2019-09-20
feat: Update API docs for image attribute

zubaira:


Philip-Larsen-Donnelly (approved):
I made a few small updates to add articles ("a" and "the") to your sentences, where I felt they were missing.

PR #261, opened 2019-10-30
fix: correct maps configuration section

amcgee:
The Maps configuration section was incorrect and out of date.

This should be backported to 2.33, 2.32, and 2.31 (I believe)


turban (approved):
Good to have this updated. We should allow GeoJSON to be entered in a future release, and allow points to be selected in a map.

amcgee:
> Good to have this updated. We should allow GeoJSON to be entered in a future release, and allow points to be selected in a map.

100% agree!

Philip-Larsen-Donnelly (approved):

PR #271, opened 2019-11-13
docs: Add docs about support of Attribute:ID scheme in events export

huzvak:
DHIS2-7070


vilkg (approved):

vilkg:
@huzvak, should this also be back-ported as the feature was implemented in all supported versions?

huzvak:
@vilkg yes, it will be. Waiting for it to be merged to the master first.

PR #285, opened 2019-11-19
fix: Remove technical architecture section

larshelge:
Removes technical architecture section which is terribly out of date.


Philip-Larsen-Donnelly (approved):
I'm fine with removing the out-of-date section, but do we plan to have a similar cohesive high-level overview of the current technical architecture somewhere? 
Will that be the job of the developer portal?

PR #286, opened 2019-11-19
Remove userGroups from user creation payload

chaseTfreeman:
reference: https://jira.dhis2.org/browse/DHIS2-3314

### DIFF
--- a/src/commonmark/en/content/developer/web-api.md
+++ b/src/commonmark/en/content/developer/web-api.md
@@ -13959,6 +13959,8 @@ support collection references etc. A simple example payload to create
 would be, the password should be sent in plain text (remember to only
 use this on a SSL enabled server) and will be encrypted on the backend:
 
+Note: that userGroups are not supported in the payload for user creation. There is a many-to-many relationship between User and UserGroup, and UserGroup is the owner of the relationship. If needed you may update UserGroup with a list of Users.
+
 ```json
 {
   "id": "Mj8balLULKp",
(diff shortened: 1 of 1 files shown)


Philip-Larsen-Donnelly:
Thanks @chaseTfreeman, but I think this PR is incorrect. 

userGroups _are_ supported on that endpoint when adding a single user.
The problem in the referenced Jira ticket is not based on this endpoint; as @vietnguyen pointed out in the ticket.

I appreciate that some clarification is needed, but I don't feel it is acceptable to hide a valid part of the API in this way.

chaseTfreeman:
@Philip-Larsen-Donnelly that is fair. Thank you for addressing the PR so quickly. However, could it still be valuable to note in the documentation that it only works when POSTing for a **single user** at a time? I can imagine that in _most_ cases, developers/users likely use the import tool or a POST request for creating more than one account at a time. At least @vietnguyen's comment about the many-to-many relationship from the ticket would be helpful for most in the documentation. I can't help but feel it is a shame that as a user trying to create many accounts I would have to go through the process of trying & failing, looking at the documentation to try and fail again _with no explanation_, and then have to find Tim's ticket on Jira "as documentation" of the actual problem. documentation should be in the docs. I'm not saying I have all the answers just that something like this should be addressed.

I'd be happy to try another PR to meet these ends.

Philip-Larsen-Donnelly:
Thanks @chaseTfreeman. I agree with you.

If you are willing to raise another PR that would be greatly appreciated! You may be in a good position to know what people are trying to do and what they need to know. Then I can ask @vietnguyen to review for technical accuracy.

PR #317, opened 2020-01-02
refactor: remove build tools from non-master branches (2.33)

Philip-Larsen-Donnelly:


Philip-Larsen-Donnelly:
Part of this is committed elsewhere. Need to review and do this again cleanly.

PR #348, opened 2020-01-14
docs: New doc structure [skip ci]

jaboto:
As the documentation for Android is being held in our specific Github repo it doesn't make sense to have it here. It creates confusion so I have deleted all the files and updated the README to explain this.


jaboto:
I am verifying something  before requesting this

PR #350, opened 2020-01-14
docs: Update dhis2_android_capture_app_INDEX.md

jaboto:
Updated to match the new folder structure in DHIS 2 Android Capture App repository

@Philip-Larsen-Donnelly  please allow this PR before nightly generation as if not it might break.


jaboto:
Hi @Philip-Larsen-Donnelly . I created the structure so we can build locally without modifying the scripts...

Philip-Larsen-Donnelly:
OK, understood, but that means you have to keep synchronising the build tools as we make changes.
You could also build locally by cloning dhis2-docs and tweaking it to point to your local INDEX, but I guess it is simpler to build independently.

PR #353, opened 2020-01-16
docs: Update dhis2_android_capture_app_INDEX.md

jaboto:
Added the line to use the about-this-guide from local repository instead of remote so it is aligned with all the other guides.


jaboto:
Will not use this in the end as the about are different. Sorry

PR #360, opened 2020-01-23
docs: remove mentions of continuous execution toggle

ghost:
Since it'll be removed for from the scheduler-app for 2.34 (https://jira.dhis2.org/browse/DHIS2-8161), this removes scheduler/continuous-execution from the web api docs.

### DIFF
--- a/src/commonmark/en/content/developer/web-api.md
+++ b/src/commonmark/en/content/developer/web-api.md
@@ -19356,11 +19356,6 @@ DHIS2 allows for scheduling of jobs of various types. Each type of job has diffe
 <td>(See list of job types)</td>
 </tr>
 <tr class="odd">
-<td>continuousExecution</td>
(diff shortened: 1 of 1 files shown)


Philip-Larsen-Donnelly:
@larshelge Has the backend been changed for this, or is 8161 just a frontend modification?

ghost:
> @larshelge Has the backend been changed for this, or is 8161 just a frontend modification?

@larshelge will know more about this than I do, but I can confirm that the backend has been changed as well. It is my understanding that the `continuousExecution` property has been removed from the jobconfigurations endpoint, and that we now have two different scheduling types (CRON and DELAY).

This is the response from https://play.dhis2.org/dev:

Screenshot 2020-02-11 at 10 10 07

And this is the response from https://play.dhis2.org/2.33.1:

Screenshot 2020-02-11 at 10 09 52

Just for some context.

larshelge (approved):

PR #387, opened 2020-02-13
feat: DHIS2-7946 - Initial documentation for the new Visualization API

maikelarabori:
Initial version for review. Please notice that I'm mentioning in a few "key" sections that the "charts" and "reportTables" API will be dropped by release 2.37. If this is not what we expect let me know so I can change it.

Let me know if you find any mistake or if you have any suggestions for improvements.

Thanks,
Maikel


Philip-Larsen-Donnelly:
Hi Maikel,

It looks fine, apart from the fact that the information about the deprecated endpoints seems a bit vague to me; in some places "will be removed" in other "should be removed" and "by 2.37" could also mean _before_ 2.37.

I would suggest we are more clear and consistent; something like:
> The "charts" and "reportTables" endpoints are deprecated in v2.34 and will be removed in v2.37.
> The combined functionality is replaced by the single "visualizations" endpoint.

I would also like to clarify how we intend to handle the api versions? I assume that currently the following are available:

```
/api/charts
/api/visualizations
...
/api/34/charts
/api/34/visualizations

```

Will `charts` and `reportTables` remain available on the unversioned API, and removed in v2.37, so that in v2.37 we have, e.g.?:
```

/api/visualizations
...
/api/37/visualizations
...
/api/36/charts
/api/36/visualizations

```

Or, do we remove the deprecated ones from the unversioned url in, say v2.35, so that it is necessary to use `/api/34` or below to access the deprecated EPs?

Some of these questions may be more for @larshelge :)

maikelarabori:
Thanks for the quick feedback @Philip-Larsen-Donnelly, I will wait for @larshelge input on the versioning thing, so I can fix all together and push a new version for another review.

Cheers,
Maikel

src/commonmark/en/content/developer/web-api.md
```diff
@@ -9531,6 +9559,619 @@ dashboard item completely:
...
+
+    /api/charts, /api/reportTables
+
+*are being replaced by*
+
+    /api/visualizations
+
+> **Note**
+>
+> New applications and clients should avoid using the "charts" and "reportTables" APIs because they are deprecated. Use the "visualizations" API instead.
```
larshelge: Use backticks instead of double quote for API resources, e.g. `charts`.

src/commonmark/en/content/developer/web-api.md
```diff
@@ -9531,6 +9559,619 @@ dashboard item completely:
...
+
+A Visualization object is composed of many attributes (some of them related to charts and others related to pivot tables), but the most important ones responsible to reflect the core information of the object are: *"id", "name", "type", "dataDimenstionItems", "columns", "rows" and "filters".*
```
larshelge: Typo: dataDimenstionItems -> dataDimensionItems

larshelge (approved):

src/commonmark/en/content/developer/web-api.md
maikelarabori: Done

src/commonmark/en/content/developer/web-api.md
maikelarabori: Done

larshelge (approved):

PR #393, opened 2020-02-21
docs(system settings): update google and bing maps api keys

varl:
Related to https://github.com/dhis2/settings-app/pull/389.
Related to https://github.com/dhis2/settings-app/pull/388.
Related to https://jira.dhis2.org/browse/DHIS2-8343.

The Google Maps API key has been moved to be co-located with the Bing Maps API key, which is also added to the docs here.


Philip-Larsen-Donnelly (commented):
It's not clear to me from this if both (Google and Bing) are required, or just one, or none?
It would be nice if we can clarify this a little better in the text.

varl:
I don't know either, it is not clear from the old docs (I just used the same strings) and I personally do not know how this works. :thinking:

Philip-Larsen-Donnelly:
I guess we need @turban to clarify this?

larshelge:
The Google API key has in fact no effect in 2.34. I think we can just keep it for a while and deprecate it as in theory external apps could still use this.

larshelge (approved):

PR #446, opened 2020-04-27
build: WIP: upgrading to material for mkdocs 5

Philip-Larsen-Donnelly:
Needs a fix for the version and language selectors on the left TOC, but otherwise looks OK.

### DIFF
--- a/src/commonmark/en/resources/css/mkdocs_dhis2.css
+++ b/src/commonmark/en/resources/css/mkdocs_dhis2.css
@@ -280,6 +280,11 @@ article {
   color: var(--grey600);
 }
 
+.md-nav__link.md-nav__link--active {
+  color: var(--blue700);
+}
+
+
 .md-nav__title{
   margin-bottom:15px;
 }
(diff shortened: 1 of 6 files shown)


src/commonmark/en/resources/css/mkdocs_dhis2.css
```diff
@@ -280,12 +280,17 @@ article {
   color: var(--grey600);
 }
 
+.md-nav__link.md-nav__link--active {
+  color: var(--blue700);
```
cooper-joe, line 284: I'd also add
`background: var(--blue100)` and `border-radius: 3px;` here to make it even clearer that it is selected.

cooper-joe: This also looks better if you add `padding: 2px;` to `.md-nav__link` class as well.

cooper-joe:
I've added some comments to make the selected section clearer.

I can't see a css fix for the scrolling-out-of-view selectors. They are inside the `.md-nav` which is a scrolling container.

Philip-Larsen-Donnelly:
Thanks @cooper-joe .

I can move the selectors. I tried putting them after the Label (e.g. "Implementer") and before the first document. Then they were visible, but it looked wrong to me.
I may need a js wizard to tell me what causes it to jump to the label and how we can make it jump to the selectors instead. My only thought was to hack it with a hidden label, then the selectors, then a second, visible, label. :/

PR #459, opened 2020-05-19
docs(settings-app): disable localised fields when default is not set

HendrikThePendric:


src/commonmark/en/content/user/system-settings.md
```diff
@@ -232,14 +232,20 @@
...
 <td><p><strong>Select language</strong></p></td>
-<td><p>Sets the language for which you can then enter translations of the following settings:</p>
-<ul>
-<li><p><strong>Application introduction</strong></p></li>
-<li><p><strong>Application title</strong></p></li>
-<li><p><strong>Application notification</strong></p></li>
-<li><p><strong>Application left-side footer</strong></p></li>
-<li><p><strong>Application right-side footer</strong></p></li>
-</ul></td>
+<td>
```
larshelge, line 235: @HendrikThePendric can we switch to markdown style tables in new docs?

HendrikThePendric: @larshelge this is part of a (much) larger table. Are you suggesting I just update this section to use markdown formatting? Or should I reformat the entire table?

larshelge: Okay merging for now.

PR #488, opened 2020-08-03
Update integration-concepts.md

armageddon:
Referenced the OpenHIM interoperability layer rather than the organisation that curates the open-source product


armageddon:
also to note, the Motech project is dead, as is the link in this page

PR #520, opened 2020-09-24
docs: update import export app docs 2.35.0

awgaan:
- Updates screenshots
- Adds tracked entity import section
- Adds tracked entity export section


src/commonmark/en/content/user/importexport-app.md
```diff
@@ -388,43 +409,76 @@ export.
...
+6.  Select a format: **XML** , **JSON** or **CSV**.
 
+7.  Click **Advanced options** if you want to adjust one or more of
+    the following settings before exporting:
+
+    * Filter by last updated date
+    * Assigned user mode
+    * Include deleted
+    * Include all attributes
+    * Data element ID scheme
```
Birkbjo, line 478: Would be nice if there was some description of these `schemes`

awgaan: An explanation of identifier schemes in general?
Or just the various commonly available scheme values (`UID`, `CODE`, `NAME` )?

src/commonmark/en/content/user/importexport-app.md
```diff
@@ -33,24 +33,16 @@ which can be found below.
...
+"Tracked Entity Instances" data), you can always view the progress of
+the import by looking at the "Job Summary" at the top of the page.
 
 ### Import Summaries
 
 <!--DHIS2-SECTION-ID:metadata_import_summaries-->
 
 On import request completion, we show import summaries above the import form.
-Any conflicts or errors are shown in a paginated table under totals.
+Any conflicts or errors are shown in a table.
```
vilkg: Is this change necessary? Can you be more specific?

vilkg (commented):
Can you update the screenshot of the app "landing page" ?

src/commonmark/en/content/user/importexport-app.md
awgaan: Removing `paginated` would've been enough, I think.
What do you think about?:
> On import request completion, we show import summaries above the
import form. Any conflicts or errors are shown in a table under the
overview summary for the import.

awgaan:
@vilkg Updated app `landing page`.
[/src/commonmark/en/content/user/resources/images/import_export/overview.png](https://github.com/dhis2/dhis2-docs/blob/0a297141ba342d494d26585279d6fd65f4f061f6/src/commonmark/en/content/user/resources/images/import_export/overview.png)

src/commonmark/en/content/user/importexport-app.md
awgaan: b113382 adds some more description about what `schemes` are and the available values for it.

vilkg:
the landing page that was added doesn't exist in the app. If it's missed intentionally, I'd say we should remove that screenshot, because having "Data import" as a landing page doesn't really add value to that section, and having a landing page that doesn't exist might look like outdated documentation. 

If joe says the landing page should be re-implemented, then the screenshot you added should stay too.

src/commonmark/en/content/user/importexport-app.md
vilkg: lgtm. I would maybe use "in the table" instead of "a table", but it's better to confirm with our native eng speakers 👍

awgaan:
- Landing page has been added to the app
- Short section about the `Job overview` page has been added to the docs

Is there anything more we need here @Philip-Larsen-Donnelly @vilkg @Birkbjo

vilkg (approved):

PR #524, opened 2020-10-05
fix: PR/PI clarification on date comparison functions

Bekkalizer:


src/commonmark/en/content/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -2206,27 +2206,27 @@ functions. The functions can be applied to data elements and attributes:
 <tr class="even">
 <td><p>d2:minutesBetween</p></td>
 <td><p>(datetime, datetime)</p></td>
-<td><p>Produces the number of minutes between two data elements/attributes of type &quot;date and time&quot;. The static datetime format is 'yyyy-MM-dd hh:mm'. Any of the arguments can be replaced with PS_EVENTDATE:(programStageUid) to compare the latest event date from a given program stage.</p></td>
+<td><p>Produces the number of minutes between two data elements/attributes of type &quot;date and time&quot;. When the first argument datetime comes before the second argument datetime, the number will be positive - in the opposite case, the number will be negative. The static datetime format is 'yyyy-MM-dd hh:mm'. Any of the arguments can be replaced with PS_EVENTDATE:(programStageUid) to compare the latest event date from a given program stage.</p></td>
```
Philip-Larsen-Donnelly, line 2209: the first always "comes before" the second, in a sense; it's less ambiguous to say "is earlier than".

You could also refer to the second argument: second argument is later than first -> number is positive 
But this might be a matter of personal taste (feels more natural to me because the "positive" number represents where the "second" datetime is relative to the first).

PR #533, opened 2020-10-22
feat: TB CS design document

YuryRog:


Philip-Larsen-Donnelly (approved):

YuryRog (commented):
Thanks. Finally adding package documentation to where it belongs!

PR #544, opened 2020-10-30
fix: complete dhis.conf part on how to disable auditing

davidhuser:
It seemed incomplete so it is copied from 2.35 docs https://docs.dhis2.org/2.35/en/dhis2_system_administration_guide/audit.html


Philip-Larsen-Donnelly (approved):
Thanks for spotting that and pushing a fix, David!

PR #557, opened 2020-11-11
docs: adds search page user documentation

paschalidi:
Have a look on this one @Bekkalizer. Adds user documentation for the search page.


src/commonmark/en/content/user/using-the-capture-app.md
```diff
@@ -409,3 +409,70 @@ In the event list you will be able to view the assignee per event. Moreover, you
...
+To execute a search now, 
+
+1. Fill in the attributes you want to search with.
+2. Click the Search by attributes button.
+
+![](resources/images/capture_app/search-by-attributes-on-scope-tetype-overview-1.png)
+
+### Fallback search
+
+Execute a full search with one of the two ways explained above. Now if the search you have made has results you will be seeing your results.
```
Bekkalizer: I guess fallback is only when searching within a program. So we should move this section within the program section, and describe it more specifically to falling back from a program to a TEI type

Bekkalizer (changes requested):

src/commonmark/en/content/user/using-the-capture-app.md
paschalidi: Right, good point

Bekkalizer (approved):

paschalidi:
@Bekkalizer I updated an image for it to have a better quality. I dont have access to merge so perhaps you could do that for us

PR #560, opened 2020-11-19
docs: updates search user manual docs for capture app

paschalidi:
Updates the user manual regarding the search page for the capture app


geethaalwan (approved):
Docs looks Ok now

src/commonmark/en/content/user/using-the-capture-app.md
```diff
@@ -429,48 +429,90 @@ The Capture app doesn't support tracker programs yet, but the tracker programs a
...
 
    ![](resources/images/capture_app/search-by-attributes-on-scope-program-overview-1.png)
+   
+3. The results of the search will be displayed as follows.
+
+   ![](resources/images/capture_app/search-by-attributes-on-scope-program-overview-2.png)
+    
+   In this list you can see the entries that match your search. For each entry you can have a total of three options. 
+    
+   a. You can choose to view the users dashboard by clicking the "View dashboard" button
```
Philip-Larsen-Donnelly, line 450: I think "user" is the wrong term here. The TE might not be a person. On the other hand "tracked entity" or "entity" are too technical. "Patient" might be more applicable than user, but it is still only relevant to people, and also only in the health domain.
This might be a question for @Bekkalizer or @Wonfrost

src/commonmark/en/content/user/using-the-capture-app.md
```diff
@@ -429,48 +429,90 @@ The Capture app doesn't support tracker programs yet, but the tracker programs a
...
 #### Fallback search
 
-Execute a full search as described above. Now if the search you have made has results you will be seeing your results.
+Execute a full search as described above. If the search you have made has results they will be displayed. However, the actual user you are searching for may be within a different program. In that case, you may want to extend the search to other programs. This is known as a fallback search.
 
-To execute a fallback search, press the button on the bottom saying "Search in all programs". 
-**Note: this is only possible in the scope of a Program.**
+To execute a fallback search, simply press the button on the bottom saying "Search in all programs". 
+
+> **Note**
```
Philip-Larsen-Donnelly, line 469: Format a note like this to make sure it is rendered nicely in our documentation.

Philip-Larsen-Donnelly (approved):
I have made a few editorial and grammatical changes.

I think it is fine as it is now, but a couple of points we should deal with in later edits:
- Should we really refer to TEs as "users"?
- In your search example you enter "Nora" but in search results you show results for "Sarah". This could be confusing.

paschalidi:
Hey Phil. Thanks for looking into those and improving by editing the text. I can open a PR replacing the Nora screenshot, I agree it can be confusing

PR #574, opened 2020-12-08
Tomcat restart after new CA scheduler job

davidhuser:
A NullPointerException without any stacktrace was thrown every time the job started, when _not_ doing a Tomcat restart. This can also be verified on play.dhis2.org/2.34.2 (but not by me as I cannot restart tomcat).

If agreed, what would be the recommended way to add the change to other branches (master, 2.35)?


Philip-Larsen-Donnelly:
Thanks @davidhuser 

I've modified your comment slightly to make is stand out in an "Important" info box (admonition).
Once I confirm the issue with Lars, I'll merge the changes.

Philip-Larsen-Donnelly:
@davidhuser 

After discussing with @larshelge, we feel this is very much a bug that should be fixed rather than documented. 
Could I ask you to raise a Jira issue for it, if you haven't already? We will deal with it with high priority.

In the meantime I will close this issue.
Thanks again!

davidhuser:
ok @Philip-Larsen-Donnelly here's the jira issue. https://jira.dhis2.org/browse/DHIS2-10106
thanks!

PR #586, opened 2021-01-05
docs: review of chapter

Philip-Larsen-Donnelly:


Philip-Larsen-Donnelly:
@benguaraldi 

I reviewed and tweaked your chapter, but Lars merged your PR before I pushed it back! :) 

So here are my changes.

Main changes:
- We cannot link to section numbers (they may change) - use relative links instead. (I am working on a refactor of the document build system, and will try to make consistent referencing via IDs in the future).
- We try to use GFM tables where possible.
- I changed your example "Implementing Partner, Improve access to medicines”	to “AIDSRelief Consortium, Improve access to medicines” - which I think is correct.
- I added some admonition block - though that is a matter of taste.

Philip-Larsen-Donnelly:
P.S. It is a nice, clearly written chapter. :)

benguaraldi:
@Philip-Larsen-Donnelly Thanks!  Looks great!

I do think that there's an advantage to the HTML tables—they look a lot clearer than the GFM tables, as they're not bolded, have the proper space around them, etc.  But all the other changes are big improvements—thank you!

benguaraldi (approved):

Philip-Larsen-Donnelly:
Hi @benguaraldi 

I agree that HTML tables often look better. The main problem is that they are not supported by our translation platform, so I am basically trying to shift the entire documentation away from them where possible; either by using GFM tables, or finding another way to represent the information. 
I understand that you are representing columns, so it makes sense as a table. One of the unfortunate issues with markdown tables is that most implementations don't support single rows. It is possible to avoid the bold by having a blank header row, and a single first row instead - but the empty header row is still rendered. We might be able to suppress that rendering as part of the new build process. If I can get that to work then I will come and attack your HTML tables again! ;)

benguaraldi:
@Philip-Larsen-Donnelly Sounds good!  Let me know if there's any way I can help!

PR #601, opened 2021-01-25
docs: adds section on user account expiration API [DHIS2-8089]

jbee:
Signed-off-by: Jan Bernitt <jaanbernitt@gmail.com>

Documentation updates in connection with https://github.com/dhis2/dhis2-core/pull/7160

PR #602, opened 2021-01-26
docs: adds a section on user account enable/disable operations

jbee:
Noticed in connection with #601 that user account `enabled` and `disabled` were not documented.


src/commonmark/en/content/developer/web-api/users.md
```diff
@@ -410,6 +410,27 @@ curl -d @replica.json "localhost/api/33/users/N3PZBUlN8vq/replica"
...
+
+<!--DHIS2-SECTION-ID:webapi_user_disable-->
+
+User accounts can be marked disabled.
+A disabled user can no longer log in.
+
+To mark a user with UID `tH7WIiIJ0O3` as disabled use:
+
+```bash
+curl -X POST "localhost/api/36/users/tH7WIiIJ0O3/disabled"
```
larshelge: Let's just use the plain API syntax without curl, we sort of stopped referring to curl in the docs.

larshelge:
Thanks for the fix @jbee . This now has a conflict, great if you can fix and then we will merge.

PR #616, opened 2021-02-04
docs: Rename app store endpoint

SferaDev:
The app store endpoint is ``/api/appHub`` and not ``/api/appStore``.


Philip-Larsen-Donnelly (approved):
Thanks for this update!

PR #632, opened 2021-03-02
docs: Hide dashboard widget app name DHIS2-9604

jimgrace:
This documents the backend change described in [DHIS2-9604](https://jira.dhis2.org/browse/DHIS2-9604). A setting in manifest.webapp can be used to suppress a dashboard widget app title from being displayed when the dashboard displays the widget.

### DIFF
--- a/src/commonmark/en/content/developer/apps.md
+++ b/src/commonmark/en/content/developer/apps.md
@@ -98,6 +98,18 @@ Among the properties are:
         }
      }
 
+  - A *settings* property is optional, and can be used on a dashboard
+    widget app to suppress showing the widget title when the widget is
+    displayed on a dashboard:
+
+<!-- end list -->
+
+    "settings": {
+        "dashboardWidget": {
+            "hideTitle": true
+        }
+    }
+
 The namespace property can be added if your app is utilizing the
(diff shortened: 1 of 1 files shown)


benguaraldi (approved):
Looks good!  Thanks, Jim!

Philip-Larsen-Donnelly:
@amcgee @deboragaleano Is this the right place for this feature to be documented. It probably doesn't harm, but I wonder if this part of the documentation is becoming deprecated in favour of content on the developer portal?

deboragaleano:
@Philip-Larsen-Donnelly sorry for my late reply! I'd say that we do want to have this type of documentation on the developer portal as well (under the App Hub section probably) but it doesn't hurt to have it here for now and I'll make sure to add this to the list of updates 👍

> @amcgee @deboragaleano Is this the right place for this feature to be documented. It probably doesn't harm, but I wonder if this part of the documentation is becoming deprecated in favour of content on the developer portal?

amcgee:
@Philip-Larsen-Donnelly I agree with @deboragaleano - app installation and manifest format are technically a feature of the DHIS2 core so they do belong here.  However it does make sense to cover this topic on the developer site and link between them as well.

PR #634, opened 2021-03-03
docs: Passive dashboard views setting DHIS2-10447

jimgrace:
See [DHIS2-10447](https://jira.dhis2.org/browse/DHIS2-10447). Adds a system setting that, when set, causes usage analytics to count passive dashboard views (such as when the user first logs in).


benguaraldi (approved):
Looks good!  Thanks, Jim!

PR #666, opened 2021-04-09
docs: add tracker error codes and its description

ameenhere:


src/commonmark/en/content/developer/web-api/new-tracker.md
```diff
@@ -44,8 +44,123 @@ A table with a full reference of error codes, messages and description:
 
 | Error Code | Error Message | Description |
 |:--|:----|:----|
-||||
-
+| E1000 | "User: `{0}`, has no write access to OrganisationUnit: `{1}`." | |
```
lucaCambi77: We could eventually just link the source file? We would need to update it manually if something is changing... It is a resource file so it is human readable in my opinion anyway

lucaCambi77 (commented):
A description of the issue would be great. Also i don't see much the point of copying/paste the errors, we could link the source file as it is already readable, but that is my opinion

src/commonmark/en/content/developer/web-api/new-tracker.md
ameenhere: Good point. I have added the description. But linking the file also needs to be updated for every version I think. A 2.36 doc should link to the 2.36 branch of the file and so on. This has to be maintained and updated in some way surely. Any thoughts @stian-sandvold ?

lucaCambi77: Yes i think you need to maintain it same way, however copying and paste can lead to more paste errors like seeing details actually changed etc while linking to a file there is no harm, you have to remember to do it though for every version as well as you have to check if errors actually changed ...

gnespolino (approved):

src/commonmark/en/content/developer/web-api/new-tracker.md
stian-sandvold: Ideally error codes does not change over time, if any, very rarely. With that in mind, it should be fine to keep a copy here, since changes should seldom occur. I mean a table like you have made, not a copy of the file itself :) Since the code should not change, the description and message should not change either.

stian-sandvold (approved):

PR #684, opened 2021-04-22
chore: prepare release 2.36

Philip-Larsen-Donnelly:


Philip-Larsen-Donnelly:
decided to move these special pages to dhis2-docs-builder

PR #691, opened 2021-04-29
docs: [DHIS2-10797] updates E1076 error 2.37

lucaCambi77:
updates E1076 error see [DHIS2-10797](https://jira.dhis2.org/browse/DHIS2-10797)

### DIFF
--- a/src/commonmark/en/content/developer/web-api/new-tracker.md
+++ b/src/commonmark/en/content/developer/web-api/new-tracker.md
@@ -890,7 +890,7 @@ There are various error codes for different error scenarios. The following table
 | E1070 | Could not find OrganisationUnit: `{0}`, linked to Enrollment. | The system could not find the OrganisationUnit specified in the Enrollment payload. |
 | E1074 | FeatureType is missing. | |
 | E1075 | Attribute: `{0}`, is missing uid. | |
-| E1076 | `{0}` `{1}` is mandatory and can''t be null | |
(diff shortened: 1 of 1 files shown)


src/commonmark/en/content/developer/web-api/new-tracker.md
```diff
@@ -890,7 +890,7 @@ There are various error codes for different error scenarios. The following table
 | E1070 | Could not find OrganisationUnit: `{0}`, linked to Enrollment. | The system could not find the OrganisationUnit specified in the Enrollment payload. |
 | E1074 | FeatureType is missing. | |
 | E1075 | Attribute: `{0}`, is missing uid. | |
-| E1076 | Attribute: `{0}`, value is null. | |
+| E1076 | `{0}` `{1}` is mandatory and can''t be null | |
```
stian-sandvold: As mentioned in the other PR; We should update the typo in this PR first, and then in the code.

gnespolino (approved):

src/commonmark/en/content/developer/web-api/new-tracker.md
Philip-Larsen-Donnelly: I don't really like the use of abbreviations like "can't" in formal documents. We should use "cannot" instead.
However, it is used in several places in this document already, so it can be corrected independently of this PR.

PR #704, opened 2021-05-05
docs: field filtering [DHIS2-10756]

gnespolino:


vilkg (approved):

developer/web-api/new-tracker.md
```diff
@@ -1312,7 +1328,7 @@ You can use a range of operators for the filtering:
 
 The `JSON` response can look like the following.
 
-Please note that field filtering (`fields=...`) support is planned but not yet implemented.
+Since version 2.37 responses can be filtered on desired fields, see [Request parameter to filter responses](#Request-parameter-to-filter-responses)
```
vilkg: I don't think we should be mentioning specific versions, since docs are already versioned

gnespolino: fair enough

stian-sandvold (approved):

PR #731, opened 2021-06-02
docs: add upgrade guide

Philip-Larsen-Donnelly:
Once this is finalised, I propose we add it after the installation guide 
in the documentation

### DIFF
--- a/sysadmin/upgrading.md
+++ b/sysadmin/upgrading.md
@@ -29,6 +29,7 @@ Regardless of whether you are upgrading or updating, the technical process is mo
 ### 1 Safeguard your data { #upgrading-safeguard-your-data }
 
 Depending on what sort of DHIS2 instance you have, and what you use it for, the first step is to make sure that you can recover any important data if anything goes wrong with the upgrade.
+
 This means performing standard system admin tasks, such as:
 
 1. Backing up your database
--- a/sysadmin/upgrading.md
+++ b/sysadmin/upgrading.md
@@ -10,7 +10,9 @@ Upgrading
 Updating
 :   Moving to a newer patch of the current DHIS2 version. e.g. from 2.35.1 to 2.35.4
 
-Upgrading typically requires planning, testing, training (for new features or interfaces), which may take significant time and effort. Updating mainly provides bug fixes without changing the functionality of the software. It is lower risk, and we advise everyone to keep their version up to date.
+**Upgrading** typically requires planning, testing, training (for new features or interfaces), which may take significant time and effort.
+
+**Updating** mainly provides bug fixes without changing the functionality of the software. It is lower risk, and we advise everyone to keep their version up to date.
 
 ## Before you begin { #upgrading-before-you-begin }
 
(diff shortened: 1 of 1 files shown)


Philip-Larsen-Donnelly:
I added this and invited the reviewers here to try to pull together a simple guide for implementers/admins, as this seems to come up often.

(Should probably add a link to dhis2.org/downloads and/or releases.dhis2.org somewhere).

sysadmin/upgrading.md
```diff
@@ -0,0 +1,53 @@
...
+
+When we talk about **Upgrading** DHIS2, we generally simply mean "moving to a newer version". However, there is an important distinction between **upgrading** and **updating**.
+
+Upgrading
+:   Moving to a newer base version of DHIS2. e.g. from 2.34 to 2.36
+
+Updating
+:   Moving to a newer patch of the current DHIS2 version. e.g. from 2.35.1 to 2.35.4
+
+Upgrading typically requires planning, testing, training (for new features or interfaces), which may take significant time and effort. Updating mainly provides bug fixes without changing the functionality of the software. It is lower risk, and we advise everyone to keep their version up to date.
```
varl: [TRIVIAL] Mostly a style suggestion. Since the point of this document is clarity, it warrants a line break.

```suggestion
**Upgrading** typically requires planning, testing, training (for new features or interfaces), which may take significant time and effort.

**Updating** mainly provides bug fixes without changing the functionality of the software. It is lower risk, and we advise everyone to keep their version up to date.
```

sysadmin/upgrading.md
```diff
@@ -0,0 +1,53 @@
...
+> If you wish to revert to an older version, you must do so with a copy of the database that was created from that older version, or previous.
+
+## Performing the upgrade { #upgrading-process }
+
+Regardless of whether you are upgrading or updating, the technical process is more-or-less identical. We will just refer to it as upgrading.
+
+### 1 Safeguard your data { #upgrading-safeguard-your-data }
+
+Depending on what sort of DHIS2 instance you have, and what you use it for, the first step is to make sure that you can recover any important data if anything goes wrong with the upgrade.
+This means performing standard system admin tasks, such as:
```
varl, line 33: [MINOR]: This line will render in the same paragraph with the former because markdown only starts a new paragraph when there is a blank line between.

I'd recommend this:

```suggestion

This means performing standard system admin tasks, such as:
```

But this would also work if you want it to be on a new line, but the same paragraph:

```
Depending on what sort of DHIS2 instance you have, and what you use it for, the first step is to make
sure that you can recover any important data if anything goes wrong with the upgrade.\
This means performing standard system admin tasks, such as:
```

Note the `\` at the end of the line before the line feed character.

varl (approved):
Looks good as a base. I've added a two presentation suggestions. Feel free to disregard them if you prefer it as is. :+1:

sysadmin/upgrading.md
Philip-Larsen-Donnelly: I usually use the "double space" convention, instead of backslash, when forcing a line break. But I am fine with your suggestion of separate block.

amcgee (approved):
LGTM!

Philip-Larsen-Donnelly:
Thanks - will publish it now. Having something will be better than nothing, I think.

PR #739, opened 2021-06-08
docs: [DHIS2-11200][DHIS2-11207] enrollment view/edit event page

simonadomnisoru:


user/using-the-capture-app.md
```diff
@@ -872,4 +872,10 @@ On the enrollment dashboard, you can view the Person Profile widget.
 
 The Person profile widget is used for viewing key attributes but not for editing. In order to edit the person profile, you must select the `Edit` button to open an edit window. 
 
+## Enrollment event view and edit page
```
larshelge, line 876: Do we want to add some text to this section?

simonadomnisoru: Hey @larshelge, 

The text is added bellow under `Reaching the enrollment event via url`, do you think I should add extra text? Thanks!

simonadomnisoru: I added the text from the section below. I hope it is better now. Thanks!

Philip-Larsen-Donnelly (approved):
Seems in line with current state of development.

PR #765, opened 2021-07-01
docs: [DHIS2-11221][DHIS2-112222] program stage event list

j-nguyenn:


src/user/using-the-capture-app.md
```diff
@@ -887,3 +887,15 @@ This is the form where you can see and edit the enrollment event details.
 In the view/edit event form you can see the stage name and icon.
 
 ![](resources/images/capture_app/enrollment-event-view-edit-header.png)
+
+
+## Program stage event list 
+
+You can reach the program stage event list by typing `/#/enrollment/stageEvents?enrollmentId={enrollmentId}&stageId={stageId}&orgUnitId={orgUnitId}&programId={programId}`, you have to specify `enrollmentId`, `stageId`, `orgUnitId` and `programId`
```
larshelge: Do we want to have API URLs in the user documentation? I suggest we keep that in the API docs.

j-nguyenn: thank you for the comment @larshelge. I just changed it to generic information and refer to [an example](https://github.com/dhis2/dhis2-docs/commit/fc005268988587e3964d65f03970bdd163c41f21) that been approved

karolinelien: Hi @huocha ,

I am just wondering if we should rather say how you reach it through the UI? (For example, "You can reach the program stage event by clicking on the "Go to full _ProgramStageName_ list" under the specific program stage in the Stages and Event list in the Enrollment overview". Maybe makes sense to also edit the example you linked to as well. Or do others have other thoughts? I am just thinking people reading the user documentation might not be the ones to create their own URLs with all these IDs.

cc @Bekkalizer @JoakimSM

j-nguyenn: hey @karolinelien , thank you for your comment.
Right now there isn't any connection established between the overview page and the full list page yet (the branch overview page is not yet validated), the only way to reach the page is through url, so I make the documentation to align with the development. 
But I agree that it could be more intuitive for user click though UI

karolinelien: That makes sense, thanks for the explanation @huocha! 
I guess it's fine then as long as we maybe do a review of the docs when these UI connections/branch overview is finalized and update them, before 2.37 release?

j-nguyenn: @karolinelien yes absolutely 👍

Philip-Larsen-Donnelly: @j-nguyenn 
Do we now have an entry point for this via the UI, so that we can update this PR?

jasminenguyennn: @Philip-Larsen-Donnelly  yes we do have a button `Go to full` however the page is not fully implemented so the button is hidden in the UI

src/user/using-the-capture-app.md
```diff
@@ -887,3 +887,15 @@ This is the form where you can see and edit the enrollment event details.
 In the view/edit event form you can see the stage name and icon.
 
 ![](resources/images/capture_app/enrollment-event-view-edit-header.png)
+
+
+## Program stage event list 
+
+You can reach the program stage event list either by clicking **Go to full {stage event name}** button in overview page or by typing in the address bar of your browser, you have to specify `enrollmentId`, `stageId`, `orgUnitId` and `programId`
```
Philip-Larsen-Donnelly: If we have a button, I would skip the bit about the address bar altogether. Keep it simple.
```suggestion
You can reach the program stage event list by clicking **Go to full {stage event name}** button in the overview page.
```

Philip-Larsen-Donnelly (approved):

PR #775, opened 2021-07-07
DHIS2-11212 Indicator widget

eirikhaugstulen:
Added documentation on the indicator widget in the enrollment overview.


src/user/using-the-capture-app.md
```diff
@@ -878,14 +878,14 @@ On the enrollment dashboard, you can view the Indicator widget.
 
 ![](resources/images/capture_app/enrollment-dash-indicator-widget-1.png)
 
-The indicator widget displays indicators from either program rules or program indicators. These indicators is a combination of keys and values.
+The indicator widget displays program indicators that are configured as "Display in form" or through program rules (Displaying text or key/value pairs in "Program indicator widget"). 
 
 #### Empty state
 If there are no indicators available for the enrollment, the widget will be shown with a text saying there are no indicators.
 If there are no program rules or indicators that can possibly display a message, the indicator will not be displayed at all.
```
karolinelien: Do you mean the indicator **widget** will not be displayed at all?

For example:
"If there are no data available for indicators related to the enrollment, the widget will be shown with a text saying there are no indicators.
If there are no program rules or program indicators configured to be displayed in the Indicator widget, the widget will not be displayed at all."

karolinelien (changes requested):
I added a commit with some suggested changes to a few sentences + a comment, let me know what you think :)

src/user/using-the-capture-app.md
```diff
@@ -872,6 +872,21 @@ On the enrollment dashboard, you can view the Person Profile widget.
 
 The Person profile widget is used for viewing key attributes but not for editing. In order to edit the person profile, you must select the `Edit` button to open an edit window. 
 
+### Indicator widget
+
+On the enrollment dashboard, you can view the Indicator widget.
```
cooper-joe: I suggest removing this line and building into the next sentence.

src/user/using-the-capture-app.md
```diff
@@ -872,6 +872,21 @@ On the enrollment dashboard, you can view the Person Profile widget.
...
+
+![](resources/images/capture_app/enrollment-dash-indicator-widget-1.png)
+
+The indicator widget displays program indicators that are configured as "Display in form" or through program rules (Displaying text or key/value pairs in "Program indicator widget").
```
cooper-joe: I suggest rewording to simplify and remove too much DHIS2 data model:
`On the enrollment dashboard, the indicator widget displays indicator text and values output related to the current dashboard.`

I don't think the user documentation needs to refer to how program indicators are set up - this seems more relevant to implementer/configuration documentation. What do you think @Bekkalizer / @karolinelien?

src/user/using-the-capture-app.md
```diff
@@ -872,6 +872,21 @@ On the enrollment dashboard, you can view the Person Profile widget.
...
+### Indicator widget
+
+On the enrollment dashboard, you can view the Indicator widget.
+
+![](resources/images/capture_app/enrollment-dash-indicator-widget-1.png)
+
+The indicator widget displays program indicators that are configured as "Display in form" or through program rules (Displaying text or key/value pairs in "Program indicator widget"). 
+
+#### Empty state
+If there are no indicators available for the enrollment, the widget will be shown with a text saying there are no indicators.
```
cooper-joe: I suggest shortening to:
`If there aren't any related indicators or indicator output for the current dashboard, the widget shows a short _empty_ message.`

src/user/using-the-capture-app.md
```diff
@@ -872,6 +872,21 @@ On the enrollment dashboard, you can view the Person Profile widget.
...
+If there are no program rules or indicators that can possibly display a message, the indicator will not be displayed at all.
```
cooper-joe: I suggest simplifying to:
`If the current dashboard can't show any indicator output (because it has no related indicators) then the widget is hidden.`

src/user/using-the-capture-app.md
```diff
@@ -872,6 +872,21 @@ On the enrollment dashboard, you can view the Person Profile widget.
...
+
+#### Legends (Optional)
```
cooper-joe: I don't think (Optional) is required here, the user doesn't have any relation to whether legends are required or not.

src/user/using-the-capture-app.md
```diff
@@ -872,6 +872,21 @@ On the enrollment dashboard, you can view the Person Profile widget.
...
+As customized in the maintenance app, program indicators can also have a `Legend` assigned to them. This will be showed as small circle to the left of the indicator value.
```
cooper-joe: I suggest simplifying to remove some DHIS2 data model references, and to focus on the main user-relevant info, which is that some indicator values will have a circle next to them:
`Some indicator values show a colored circle next to the value. The colored circle shows the related legend color for that indicator value. Colored legend circles are only shown for indicator values that have them set up.`

cooper-joe (changes requested):

eirikhaugstulen:
Thanks everyone!
Implemented the changes suggested by @cooper-joe, but feel free to give it another look if you still think something should be rephrased 👍

cooper-joe (approved):

simonadomnisoru (approved):

src/user/using-the-capture-app.md
karolinelien: To me the user documentation is a mix of user and configuration documentation - for example the chapter on "configuring the system", specifically on program rules here: https://docs.dhis2.org/en/use/user-guides/dhis-core-version-236/configuring-the-system/programs.html#configure_program_rule - maybe we should specify it more there instead (Just says "display widget" - could say this refers to Indicator widget in Capture app) - I just want to make sure the user/configurer can read about what to do to actually get the indicators to display in the widget _somewhere_ in the docs :)

Philip-Larsen-Donnelly: I think the current level of detail is fine for the user documentation related to DHIS2-11212.

Philip-Larsen-Donnelly (approved):

PR #790, opened 2021-08-11
docs: [DHIS2-10222] add docs for stages and events list

j-nguyenn:
## [Feature](https://jira.dhis2.org/browse/DHIS2-10222)

This ticket is to add documentation for Program stage list on Enrollment page. 
Navigation between overview page were implemented and documented (The buttons `Go to xxx` and `New xxx` )


Philip-Larsen-Donnelly (changes requested):
The text looks fine, but I have a couple of comments on the markdown.

1. Please separate blocks with a blank line, to avoid parsing ambiguities.
e.g.
- blank line after headings
- blank lines before and after images that are not "inline".

2. When referring to App sections or buttons the classic approach was to put them in bold.
(We used clarify this at the start of most documents, but the new format has removed that and it's only kept in a few places, like [here](https://docs.dhis2.org/en/implement/configuring-the-android-app/about-this-guide.html)).
This Capture app chapter uses several different styles: **Bold**, "Quotes", and in this PR `code`. I think both **Bold** and "Quotes" are preferable when referring to sections or buttons on the app interface.

j-nguyenn:
Thank you @Philip-Larsen-Donnelly for the feedback, I changed and updated the branch as requested.

Philip-Larsen-Donnelly:
Thanks for the updates @huocha !

Looking a little closer, I actually have a question about the "Reset" button. Is this something that came from a design by @cooper-joe (and is it used elsewhere in a similar context)? The reason I ask is that, from a user perspective, if I see a "Reset" button, I am worried I might remove data. If I understand the documentation correctly, this is just for resetting the view? If so, maybe we should be more explicit by labeling the button something like "Reset view"?

cooper-joe:
@Philip-Larsen-Donnelly is right, and this is how the button is designed and documented in the specification, so the implementation isn't right.
[Direct link to the section in the specification](https://docs.google.com/document/d/19I506o6P9ODf7M0TfaZnP0N98wn_eYuNmBQ_T21wm7Y/edit#heading=h.gf61owqvf6b9).

"_When the list has been expanded with a click of “Show 5 more”, or the sorting is changed - a (ui secondary, small) button appears last in the button group with the label Reset list. Clicking the Reset list button returns the table back to the default sorting and initial 5 events._"

j-nguyenn:
@Philip-Larsen-Donnelly the functionality is documented [here](https://docs.google.com/document/d/19I506o6P9ODf7M0TfaZnP0N98wn_eYuNmBQ_T21wm7Y/edit) 
maybe the wording I used should be `returns the table back to the default sorting and initial 5 events` instead of `to initial state`. 

The functionality is correctly implemented

Philip-Larsen-Donnelly:
@huocha Thanks. I think the documentation is fine, the issue is only the button label - should be "Reset list" as per the design spec. (and the screenshot should reflect that too, of course).

j-nguyenn:
@Philip-Larsen-Donnelly this should be updated

Philip-Larsen-Donnelly:
Thanks @huocha !

Lastly... :grimacing: your screenshots are quite high resolution - not really following our guidelines https://github.com/dhis2/dhis2-docs#taking-screenshots.

Philip-Larsen-Donnelly (approved):

PR #796, opened 2021-08-20
doc: reviewed section on using the DHIS2 web API with R

INyabuto:
- updated reference for the two packages currently used in the section. 
- reviewed the first example for clarity on the packages loaded.
- Included some missing parts on the example just to that ensure the code works.


Philip-Larsen-Donnelly (approved):
Thanks @INyabuto !

PR #815, opened 2021-09-02
docs: A little more information about how SQL views work

benguaraldi:
One of the PEPFAR developers was having difficulty figuring out how to make SQL views that called other SQL views, which caused me to add a little detail to the relevant section of the user manual.


jimgrace (approved):

Bekkalizer (approved):
LGTM!

PR #823, opened 2021-09-10
docs: params for image api queries

jaboto:
While querying resources keywords must be in capital letters otherwise resulting in an exception from the server.


Bekkalizer (approved):
LGTM!

PR #839, opened 2021-10-04
update example, because it is not working

dayures:
this is not the way for querying for ID1 or ID2.
The current expression searchs for ID1 and ID2, that always returns empty (except if ID1 = ID2)


Philip-Larsen-Donnelly (approved):

dayures:
@Philip-Larsen-Donnelly should this PR be merged in the main branch (not only 2.36) ?

dayures:
@Philip-Larsen-Donnelly I see that this example is still appearing in the master documentation

PR #843, opened 2021-10-07
docs: Clarifies the relationship between externalAccess and publicAccess

benguaraldi:
As discussed in [Jira 11848](https://jira.dhis2.org/browse/DHIS2-11848), it is possible to set up externalAccess and publicAccess to provide access to objects only when no user is logged in (and prevent access when a user is logged in).

This seemed like a good thing to mention in the documentation, so I added it everywhere that externalAccess is mentioned.

### DIFF
--- a/src/developer/web-api/sharing.md
+++ b/src/developer/web-api/sharing.md
@@ -78,6 +78,13 @@ curl -d @sharing.json "localhost/api/33/sharing?type=dataElement&id=fbfJHSPpUQD"
   -H "Content-Type:application/json" -u admin:district
 ```
 
+> **Note**
+>
+> It is possible to create illogical sharing combinations. For
+> instance, if `externalAccess` is set to `true` but `publicAccess` is
+> set to `--------`, then users will have access to the object 
+> ***only when they are logged out.***
+
 ## Cascade Sharing for Dashboard
 
 ### Overview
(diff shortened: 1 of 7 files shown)


src/developer/web-api/sharing.md
```diff
@@ -78,6 +78,13 @@ curl -d @sharing.json "localhost/api/33/sharing?type=dataElement&id=fbfJHSPpUQD"
   -H "Content-Type:application/json" -u admin:district
 ```
 
+> **Note**
+>
+> It is possible to create illogical sharing combinations. For
```
jimgrace: I know I used the word illogical for this combination. But in user documentation I would avoid this word as a kindness both to the product and to users who might find some strange use for this combination. Also it is no more illogical than unix file sharing permissions where you could give everyone more access than the user or group owner.

Instead of "It is possible to create illogical sharing combinations." I would say "It is possible to create sharing combinations that you may not intend."

jimgrace (changes requested):
One suggested change. The rest looks great. Thanks!

src/developer/web-api/sharing.md
benguaraldi: Yeah, I agree—illogical was too judgy.  I've changed it to "surprising" so it now reads:
> It is possible to create surprising sharing combinations. For instance, if `externalAccess` is set to `true` but `publicAccess` is set to `--------`, then users will have access to the object  only when they are logged out.

Does that work?

jimgrace: That's great.

jimgrace (approved):

PR #875, opened 2021-11-01
chore: Add import CSV events [DHIS2-10837]

enricocolasante:


src/developer/web-api/new-tracker.md
```diff
@@ -579,6 +578,21 @@ Examples of the **ASYNC** response is shown below. For **SYNC** response, look a
 }
 ```
 
+### CSV Events payload
+
+In order to maintain compatibility with older versions of tracker, the API allow to import events using the CSV format.
```
enricocolasante:
```suggestion
In order to maintain compatibility with older versions of tracker, the API allows to import events using the CSV format.
```

PR #905, opened 2022-01-05
docs: adds section on datastore field filtering [DHIS2-12154]

jbee:
Docs for https://github.com/dhis2/dhis2-core/pull/9526


jbee:
Is replaced by #931 which extends the addition made in this PR.

PR #909, opened 2022-01-10
chore: add documentation for Vector tile basemaps

jenniferarnesen:


src/user/configure-metadata.md
```diff
@@ -4571,6 +4574,8 @@ Table: External map layer objects in the Maintenance app
         be added from the Add Layer selection and placed anywhere above
         the base map.
 
+    Note that Vector Style layers can only be added as a Basemap.
```
turban: "basemap" in lowercase. We should also change from "base map" to "basemap" in line 4575.

src/user/configure-metadata.md
```diff
@@ -4548,6 +4547,10 @@ Table: External map layer objects in the Maintenance app
 
       - XYZ tiles (can also be used for WMTS)
 
+      - Vector tiles (Vector Style)
+
```
turban, line 4551: Add: "The map service URL should be to a JSON document that follows the [Mapbox GL Style Spec](https://docs.mapbox.com/mapbox-gl-js/style-spec/)."

PR #911, opened 2022-01-12
Docs: fix PR, number must be enclosed in 's.quote'

gassims:
line 1352, program rule d2:addDays example uses a number without enclosing it between a single qoute which generates an error in the validation engine, so instead of:
`d2:addDays(#{lastMenstrualDate},283)`
We'll use:
`d2:addDays(#{lastMenstrualDate},'283')`

Thanks!


Philip-Larsen-Donnelly:
@Bekkalizer Is this true that the number of days must be provided in single quotes (i.e. as a string)?

gassims:
> @Bekkalizer Is this true that the number of days must be provided in single quotes (i.e. as a string)?

@Philip-Larsen-Donnelly I actually tried both options and the only one that turned valid was the one with a single quotes. I hope @Bekkalizer would confirm and sorry if I'm wrong. (:

PR #913, opened 2022-01-13
docs: [DHIS2-11862] add suggested date rules

jasminenguyennn:
### Docs
Add suggested date rules on schedule event page

### DIFF
--- a/src/user/using-the-capture-app.md
+++ b/src/user/using-the-capture-app.md
@@ -1058,7 +1058,15 @@ This is the form where you can modify the event details before saving. In the he
 ### Schedule event widget form 
 Instead of reporting an event the user can select to schedule an event later. This is done with a scheduled date. The dialog will open with a suggested scheduled date, and this date is determined by a set of rules from program stage configuration and program configuration.
 
+- if program stage has **Next schedule date** then the suggested date is the most recent next schedule date
+- if program has **Standard interval days** then the suggested date is calculated by the most recent event date plus the standard interval days value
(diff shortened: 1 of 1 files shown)


src/user/using-the-capture-app.md
```diff
@@ -1058,8 +1058,16 @@ This is the form where you can modify the event details before saving. In the he
 ### Schedule event widget form 
 Instead of reporting an event the user can select to schedule an event later. This is done with a scheduled date. The dialog will open with a suggested scheduled date, and this date is determined by a set of rules from program stage configuration and program configuration.
 
+- if program stage has **Next schedule date** then the suggested date is the most recent next schedule date
+- if program has **Standard interval days** then the suggested date is calculated by the most recent event date plus the standard interval days value
+- if no value found after these 2 steps, then the suggested date will be defined by enrollment date and incident date. In case the option **Generate events based on enrollment date** is checked in Maintenance app, the suggested date is calculated by the enrollment date plus the value of min days from the start. In case the option **Show incident date** is checked, the suggested date is calculated by the incident date plus the value of min days from the start.
```
karolinelien:
```suggestion
- if no value is found after these 2 steps, then the suggested date will be defined by enrollment date and incident date. In case the option **Generate events based on enrollment date** is checked in the Maintenance app, the suggested date is calculated by the enrollment date plus the value of minimum days from the start. In case the option **Show incident date** is checked, the suggested date is calculated by the incident date plus the value of minimum days from the start.
```

karolinelien: Where do you set "Minimum days from the start"? Do you mean "Scheduled days from start"?

karolinelien (changes requested):

PR #915, opened 2022-01-13
feat: [DHIS2-11953] Docs on quick actions widget

eirikhaugstulen:
Added docs on the quick actions-widget in the enrollment overview.


src/user/using-the-capture-app.md
```diff
@@ -848,6 +848,13 @@ When you deselect the enrollment you see the following
 
 ![](resources/images/capture_app/enrollment-dash-08.png)
 
+
+###Quick actions
+
+The quick actions-widget contains shortcuts to the most common actions performed by the user.
```
cooper-joe: Two points here:
- This is user documentation, so I don't think we should refer to `the user`. 
- I don't think the hypen in -widget is necessary.

I suggest:
_The quick actions widget offers shortcuts for frequently used actions for the current enrollment._

cooper-joe (changes requested):

src/user/using-the-capture-app.md
karolinelien: I like Joe's suggestion!

eirikhaugstulen:
Thanks @cooper-joe & @karolinelien! 
Should be updated now.

cooper-joe (approved):

PR #923, opened 2022-01-21
docs: [DHIS2-10946] Tracked entity instance edit modal

simonadomnisoru:


karolinelien:
Looks good to me! Should we say anything about validations or anything like that? Or how the key attributes are selected (display in list). Maybe inferred.

src/user/using-the-capture-app.md
```diff
@@ -921,14 +921,15 @@ The enrollment comment widget displays comments and allows addition of comments,
...
-#### Person profile widget
+#### Tracked entity instance profile widget
 
-On the enrollment dashboard, you can view the Person Profile widget.
+On the enrollment dashboard, you can view the tracked entity instance profile widget. Inside the profile widget you can view the key attributes values. 
 
 ![](resources/images/capture_app/enrollment-dash-tei-profile-widget.png)
 
-The Person profile widget is used for viewing key attributes but not for editing. In order to edit the person profile, you must select the `Edit` button to open an edit window. 
+In order to edit the tracked entity instance profile, you must click the `Edit` button. The edit modal will open and you can modify and save the attributes values.
```
cooper-joe: I think this statement could be simplified and shortened. Also, elsewhere in the document bold text is used for button labels, rather than `code`. My suggestion:
"Click the **Edit** button to make changes to the tracked entity instance profile. Editing the profile opens a dialog where the profile attributes can be changed."

cooper-joe (commented):
I added a suggestion to simplify part of the text.

> Should we say anything about validations or anything like that? Or how the key attributes are selected (display in list).

@karolinelien I think that might go beyond the scope of this document, as this is user documentation. Seems like the kind of information that should be included in Implementer documentation though for sure.

cooper-joe (approved):

PR #940, opened 2022-02-15
docs: Adds ADX options for Import-Export app [DHIS2-4978]

benguaraldi:
Adds the ADX options for exporting in the Import-Export app to the documentation.

See https://jira.dhis2.org/browse/DHIS2-4978 for more details.


jimgrace (approved):
That was good, careful editing. Thanks!

PR #946, opened 2022-02-25
docs: custom periods [DHIS2-12323]

gnespolino:


src/developer/web-api/overview.md
```diff
@@ -526,6 +526,49 @@ for creating dynamic reports. The available relative period values are:
     LAST_3_MONTHS, LAST_6_BIMONTHS, LAST_4_QUARTERS, LAST_2_SIXMONTHS, THIS_FINANCIAL_YEAR,
     LAST_FINANCIAL_YEAR, LAST_5_FINANCIAL_YEARS
 
+### Custom date periods { #webapi_date_custom_date_periods }
+
+Analytics `query` resources support extra parameters to express periods.
+
+Default `pe` dimension fallbacks to:
+
+- `executionDate` for `/analytics/events/query`
```
vilkg: I suggest to say "eventDate" here to be consistent with the query param description and the data model. executiondate is db column name, but it's called eventDate in the API.

gnespolino: true, thank you

vilkg (approved):

PR #974, opened 2022-03-19
docs: predictor job scheduling writeup

jimgrace:


src/user/scheduling.md
```diff
@@ -319,3 +319,99 @@ Some aspects of the meta data synchronization feature to be aware of:
 
 - You can see the time of last successful synchronization with remote
   server in the scheduling screen next to the "Last success" label.
+
+### Predictor { #scheduling_predictor } 
+
+This runs selected predictors and/or predictor groups.
+
+The relative start and relative end parameters determine the periods in
+which data will be predicted, relative to the date on which the
```
benguaraldi: The three "relatives" are accurate, but they make it harder for me to read.

Maybe skip the second one and change the third to "corresponding", i.e.:
 > The relative start and end parameters determine the periods in which data will be predicted, corresponding to the date on which the predictor job is run:

src/user/scheduling.md
```diff
@@ -319,3 +319,99 @@ Some aspects of the meta data synchronization feature to be aware of:
...
+think carefully about when you want predictions for a period to start
+being made, and when you want them to stop being made. Then you
+need to compute the appropriate relative start
+and end dates.
+
+Examples:
+
+1. A predictor is computed from weekly data of the same week. (No past
+sampled data is used.) After the week ends on Sunday,
+you expect the weekly data to be entered on the Monday and Tuesday.
```
benguaraldi: Maybe "the following Monday and Tuesday" is clearer?

src/user/scheduling.md
```diff
@@ -319,3 +319,99 @@ Some aspects of the meta data synchronization feature to be aware of:
...
+predictions will be made, especially if your predictor job is set to
+run daily or more frequently. Before you set these values, you should
+think carefully about when you want predictions for a period to start
+being made, and when you want them to stop being made. Then you
+need to compute the appropriate relative start
+and end dates.
+
+Examples:
+
+1. A predictor is computed from weekly data of the same week. (No past
```
benguaraldi: "A predictor is computed from weekly data of the same week" is slightly awkward, but I'm not sure how to fix it.  "A predictor is computed from weekly data that is all entered within one week"?

src/user/scheduling.md
```diff
@@ -319,3 +319,99 @@ Some aspects of the meta data synchronization feature to be aware of:
...
+
+    - On Thursday of the following week, the period started 10 days
+    before and ended 3 days before. Predictions are made because -10 to -3
+    are within the range -10 to -2.
+
+    - After Thursday, the previous week started more than
+    10 days before, so no predictions are made.
+
+    - Predictions are made only on Wednesday and Thursday. On Friday through
+    Tuesday, no predictions are made (and the job finishes very quickly).
```
benguaraldi, line 382: The above is tricky and very well explained—good job!

src/user/scheduling.md
benguaraldi: Possibly add the word "Problem:" or "Situation:" in bold before "A predictor is computed" and then "Solution:" before "For a job running"?

src/user/scheduling.md
```diff
@@ -319,3 +319,99 @@ Some aspects of the meta data synchronization feature to be aware of:
...
+
+2. A predictor is used to forecast a limit (average plus twice the standard deviation)
+for expected non-seasonally varying disease cases based on data from the
+previous five weeks. Weeks are Monday through Sunday. Predictions should start
+being made from the previous Tuesday, using available data at that time, and
+continue being made through Tuesday of the week that the predictions are being
+made for (by which it is assumed that the previous week's data is final).
+
+    Define relative start as -1 and relative end as 12.
```
benguaraldi: Possibly add the word "Problem:" or "Situation:" in bold before "A predictor is used" and then "Solution:" before "Define relative start"?

(Only if Problem and Solution are used in example 1, of course.)

benguaraldi: Should it be specified that the predictions are daily for this as well?  I noticed you said that in example 1 but not here, but I'm not sure if there's a difference I'm missing.

benguaraldi (approved):
Looks great!  A few comments, which I hope are helpful, but all could be safely ignored.

src/user/scheduling.md
jimgrace: Yeah, I have a lot of relatives. Fixed.

src/user/scheduling.md
jimgrace: Changed to "After the week ends on Sunday, you expect the data to be entered in the following two days: Monday and Tuesday."

src/user/scheduling.md
jimgrace: Changed to: "A predictor uses data from the same week as the predicted value. (No past sampled data is used.)"

Good idea. I'm using **Requirement:** and **Solution:**

src/user/scheduling.md
jimgrace: Thanks, I think the whole thing is rather tricky. I spent a lot of time changing words between "before", "earlier", "less than", "from", "within", "between", etc. They all mean the same thing (usually), but some read more easily than others. I appreciate your careful reading and good suggestions!

src/user/scheduling.md
jimgrace: Added **Requirement:** and **Solution:**

I meant the same as before, thinking the user would catch on. But I've now added "For a job running daily or more frequently," to be clear. Good idea.

benguaraldi: Looks good!

src/user/scheduling.md
benguaraldi: I agree that the whole thing is tricky—folks are likely to have to read this a few times and play around with it to understand it.  But everything they need to understand it is here—again, good job!

src/user/scheduling.md
benguaraldi: Looks good!

src/user/scheduling.md
benguaraldi: This works fine, but I would suggest, "the following two days (Monday and Tuesday)."

src/user/scheduling.md
benguaraldi: Looks good, thanks!

benguaraldi (approved):

src/user/scheduling.md
jimgrace: Perfect, thanks!

PR #981, opened 2022-03-24
docs: better and updated job parameter tables

jbee:
The existing table was hard to maintain and more importantly to read. 
I split the one table into many tables, one for each job.

Also I checked if the parameters were still correct and updated (added) missing parameters that were so far not documented.

This PR so far contains only those job types that were present in the table.
I think some job types are completely missing and should be added.


Philip-Larsen-Donnelly (approved):
Great improvement! Thanks!

PR #988, opened 2022-03-29
indexing of repeatable program stages

d-bernat:
Documenation to indexing of repeatable program stages, please see https://jira.dhis2.org/browse/DHIS2-11190


jimgrace (changes requested):
This looks great, thanks! (It should have been approved a long time ago, and it should be backported to 2.38 when approved.) We should also document the .stageOffset() function as it applies for program indicators.

larshelge (approved):

PR #990, opened 2022-04-11
Update i18n.md

chaseTfreeman:
Updated resource endpoint for db locales and add required parameters to POST db locales


gassims (approved):
Related Community of Practice topic:
https://community.dhis2.org/t/is-it-possible-to-change-the-name-of-a-database-locale/46853

Confirmed. Using the endpoint /api/33/locales/db for POST requests doesn't work whereas /api/locales/dbLocales POST requests work if required parameters are submitted.

PR #1006, opened 2022-05-19
docs: adds GeoJSON import documentation [DHIS2-7171]

jbee:
Documentation for changes in new GeoJSON features from:
* https://github.com/dhis2/dhis2-core/pull/10838 
* https://github.com/dhis2/dhis2-core/pull/10863
* https://github.com/dhis2/dhis2-core/pull/10875
* https://github.com/dhis2/dhis2-core/pull/10963

### DIFF
--- a/src/developer/web-api/metadata.md
+++ b/src/developer/web-api/metadata.md
@@ -45,7 +45,8 @@ Table: Scheme Values
 | ATTRIBUTE:ID | Match on metadata attribute, this attribute needs to be assigned to the type you are matching on, and also that the unique property is set to *true*. The main usage of this is also to exchange data with external systems, it has some advantages over *CODE* since multiple attributes can be added, so it can be used to synchronize with more than one system. |
 
 Note that identifier schemes is not an independent feature but needs to
-be used in combination with resources such as data value import and metadata import.
+be used in combination with resources such as data value import, metadata import and
+GeoJson import.
 
 As an example, to specify CODE as the general id scheme and override
 with UID for organisation unit id scheme you can use these query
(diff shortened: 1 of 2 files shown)


src/developer/web-api/metadata.md
```diff
@@ -1071,6 +1072,7 @@ ignored:
 }
 ```
 
+!INCLUDE "geo-json.md"
```
jbee, line 1075: @Philip-Larsen-Donnelly I think this is the first time someone actually uses includes. Maybe you can check if this actually works? Thanks!

Philip-Larsen-Donnelly: Includes should still work. I can test this. 
We have generally moved away from includes as it breaks the concept of providing an edit button for each page in the documentation portal. (The edit button will take the user to the file in the docs git repo - but will not follow includes).

jbee: That is an unfortunate complication. I thought these files get so large includes would help to break this up and make the edit in an IDE manageable. So at the same time this breaks the portals edit.

Philip-Larsen-Donnelly: Yeah, the original idea of the includes was to make it more manageable, but I feel that the accessibility from the portal is pretty important; and we can consider breaking sections into multiple pages if they become too large and unmanageable.

jbee: Yes, I think breaking the metadata.md in multiple pages is the best. That way the imported one can just become one of the pages in the sequence of pages.

larshelge (approved):

PR #1087, opened 2022-10-11
feat: Support for new flag skipOrgUnitOwnership [DHIS2-13873]

maikelarabori:
Documenting the new query parameter flag that can be set in the Analytics Export job.
The endpoint related to that job is `/api/resourceTables/analytics`

Example of usage:
`/api/resourceTables/analytics?skipOrgUnitOwnership=true`


jimgrace (approved):
Looks perfect, thanks!

Philip-Larsen-Donnelly (approved):

PR #1088, opened 2022-10-11
docs(aggregate data entry): add initial docs for new app

Mohammer5:


src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+The **Data Entry** app (beta) is where you manually enter aggregated data in
+DHIS2. You register data for an organisation unit, a period, and a set
+of data elements (data set) at a time. A data set often corresponds to a
+paper-based data collection tool. You configure the data sets in the
+**Maintenance** app.
+
+> **Note**
+> 
+> If a data set has both a section form and a custom form, the system
+> displays the custom form during data entry. Users who enter data can't
```
Philip-Larsen-Donnelly: Avoid using "informal" abbreviations in formal documentation:
```suggestion
> displays the custom form during data entry. Users who enter data cannot
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+> 3.  Default form
+> 
+> Mobile devices do not support custom forms. In mobile-based data entry
+> the order of display preference is:
+> 
+> 1.  Section form (if it exists)
+> 
+> 2.  Default form
+
+When you close an organisation unit, you can't register or edit data to
```
Philip-Larsen-Donnelly:
```suggestion
When you close an organisation unit, you can no longer register or edit data to
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+
+![](resources/images/data_entry_beta/selecting_a_form.png)
+
+In order to show a form, the following information is required:
+
+1.  Open the **Data Entry** app (beta).
+
+1.  A data set
+
+    As selectable the organisation unit and period depend on the data set, the
```
Philip-Larsen-Donnelly:
```suggestion
    As the organisation unit and period depend on the data set, the
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+    The available periods are controlled by the period type of the data
+    set (reporting frequency). You can jump a year back or forward by
+    clicking **Prev year** or **Next year**.
+
+1.  Attribute options (in case the data set has an attribute option combo)
+
+    The attribute option combo also depends on the period and organisation
+    unit, so the input will be hidden until all required values have been
+    selected. If a data set does not have an attribute option combo, then no
+    selector will appear.
```
Philip-Larsen-Donnelly: I'm not sure that we should talk about "attribute option combo's" in end user documentation - that's quite a technical description. 
I think we need to talk simply about "options"?

```suggestion
1.  Additional options

    The available options also depend on the period and organisation
    unit, so the input will be hidden until all required values have been
    selected. If a data set does not have an additional options, then no
    selector will appear.
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+
+1.  Select a form (see [Selecting a form](#aggregate_data_entry_app.selecting_a_form))
+
+1.  Enter data in the data entry form.
+    
+      - A green field with a green triangle in the top right corner means that
+        the system has saved the value.
+    
+      - A grey field with a "not allowed" mouse cursor means that the field is
+        disabled and you can't enter a value. The keyboard cursor will
```
Philip-Larsen-Donnelly:
```suggestion
        disabled and you cannot enter a value. The keyboard cursor will
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+        automatically jump to the next open field.
+    
+      - To move to the next field, press the Tab key or the Down Arrow
+        key.
+    
+      - To move back to the previous field, press Shift+Tab or the Up
+        Arrow key.
+    
+      - If you type in an invalid value, for example a character in a field
+        that only accepts numeric values, you'll get a tooltip that explains
```
Philip-Larsen-Donnelly:
```suggestion
        that only accepts numeric values, you will get a tooltip that explains
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+        the problem and the field will be coloured red until you have corrected
+        the value.
+    
+      - If you have defined a minimum maximum value range for the field and you
+        enter a value that is outside this range, you'll get a pop-up message
```
Philip-Larsen-Donnelly:
```suggestion
        enter a value that is outside this range, you will get a pop-up message
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+        that says the value is out of range. The field will be coloured red
+        until you have corrected the value.
+
+      - When clicking on "Options" and then on "Help", you will see the cell
+        reference which contains a legend of the different cell states.
+
+6.  When you've filled in the form, click **Run validation** in the bottom
```
Philip-Larsen-Donnelly:
```suggestion
6.  When you have filled in the form, click **Run validation** in the bottom
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+    left corner in the bottom bar.
+    
+    All validation rules which involves data elements in the current
```
Philip-Larsen-Donnelly:
```suggestion
    All validation rules related to data elements in the current
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+    ![](resources/images/data_entry_beta/validation_rule_result.png)
+
+7.  (Optional) Correct validation violations.
+    
+    > **Note**
+    > 
+    > Zero (0) will delete the value if the data element has been
+    > configured to not store zeros.
+
+8.  When you've corrected errors and you're done with data entry, click
```
Philip-Larsen-Donnelly:
```suggestion
8.  When you have corrected errors and you have finished entering data for the entire form, click
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+
+Each field in a form (a particular data value) can have additional information:
+
+*   Basic information (description, code, data element ID, category option
+    combo ID & last time changed, whether the data value has been marked for
+    follow-up or not)
+
+*   A comment
+
+*   Min- anx max limits
```
Philip-Larsen-Donnelly:
```suggestion
*   Min (minimum) and max (maximum) limits
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+
+*   History
+
+*   Audit log
+
+These information can be viewed by opening the details for a particular field.
```
Philip-Larsen-Donnelly:
```suggestion
This information can be viewed by opening the details for a particular field.
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+
+1.  Open the **Data Entry** app (beta).
+
+1.  Open an existing data entry form.
+
+1.  Focus on a field.
+
+1.  Open the details sidebar (by clicking Ctrl+Enter, Cmd+Enter, double
+    clicking the field or clicking the "Show details" button in the bottom bar)
```
Philip-Larsen-Donnelly, line 155:
```suggestion
    clicking the field or clicking the "View details" button in the bottom bar)
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+
+3.  Enter **Min limit** and **Max limit**.
+
+4.  Click **Save**.
+
+    ![](resources/images/data_entry_beta/save_new_limits.png)
+    
+    If values don't fall within the new value range the next time you enter
+    data, the data entry cell will appear with a red background.
+
```
Philip-Larsen-Donnelly, line 210:
```suggestion
    If values don't fall within the new value range the next time you enter
    data, the data entry cell will appear with a red background.

    > **Note**
    >
    > Only users with the relevant privileges are able to edit the value ranges.
    
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+4.  Click **History**.
+
+![](resources/images/data_entry_beta/history.png)
+
+## Display a data value's audit log { #aggregate_data_entry_app.display_data_value_audit_log } 
+
+The audit log allows you to view other data values which have been entered
+prior to the current value. The audit trail also shows when the data value was
+altered and which user who made the changes.
```
Philip-Larsen-Donnelly:
```suggestion
altered and which user made the changes.
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+## Enter data offline { #aggregate_data_entry_app.enter_data_offline } 
+
+The **Data Entry** app (beta) works even if you don't have a stable Internet
+connection during data entry. When you don't have an internet
+connection, the data you enter is saved to your local computer. When the
+Internet connection is back, the app will push the data to the server.
+
+> **Note**
+> 
+> To use this functionality, you must login to the server while you've
```
Philip-Larsen-Donnelly:
```suggestion
> To use this functionality, you must login to the server while you have
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+    as normal.
+
+-   Once you have entered data and the app detects that the Internet connection
+    is back, the data will be sent automatically:
+
+## Multi-organisation unit data entry { #aggregate_data_entry_app.data_entry_multiple_organisation_units } 
+
+The **Data Entry** app (beta) currently does not support multi-organisation
+units!
+
```
Philip-Larsen-Donnelly, line 277: Do we need to say this? Why not remove altogether. 

```suggestion
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,283 @@
...
+## See also { #aggregate_data_entry_app.data_entry_app_see_also } 
+
+  - [Control data
+    quality](https://docs.dhis2.org/master/en/user/html/control_data_quality.html)
+
+  - [Manage data sets and data entry
+    forms](https://docs.dhis2.org/master/en/user/html/manage_data_set.html)
+
+  - [Using the Maintenance
+    app](https://docs.dhis2.org/master/en/user/html/maintenance_app.html)
```
Philip-Larsen-Donnelly: Refer to unique anchors directly rather than full URLs which may change.
(see https://docs.dhis2.org/en/implement/support-and-documentation/dhis2-documentation-guide.html#name_of_section)

```suggestion
  - [Manage data sets and data entry
    forms](#manage_data_set)
    
  - [Control data
    quality](#control_data_quality)
```

Philip-Larsen-Donnelly (commented):
Some review comments

src/user/using-the-aggregate-data-entry-app.md
Mohammer5: The reason I included it is: If a user is wondering how this works in the new app and don't find any information about it, they could be confused.

I can remove it if you think we shouldn't put it in there

src/user/using-the-aggregate-data-entry-app.md
Mohammer5: The button's label says "Show details"

src/user/using-the-aggregate-data-entry-app.md
Mohammer5: The app also supports option sets, which are category options for a particular field (I think..). The top bar also has a dropdown button with the label "Options", so that word can have multiple meanings.

If you think it's still fine to simply refer to "options" when talking about selecting a form, I can commit the suggestion

Mohammer5:
@Philip-Larsen-Donnelly I've copied a lot of texts from "[Using the Data Entry app](https://github.com/dhis2/dhis2-docs/blob/master/src/user/using-the-data-entry-app.md)". Should I apply your suggestion to that file as well?

src/user/using-the-aggregate-data-entry-app.md
Philip-Larsen-Donnelly: To be honest I wasn't familiar with this feature, but I see it is in the old app. 
In that case, I agree it may make sense to mention it as you have done.

src/user/using-the-aggregate-data-entry-app.md
Philip-Larsen-Donnelly: Perhaps we can run this by @cooper-joe ?

The point is that I think end users shouldn't have to care what is an attribute option, or a category option, or whatever; all they need to know is that there are additional categories* that they have to select.

*from a purely end-user point of view I would probably prefer to refer to "additional categories" than "additional options"; so one modification of my suggestion would be to change `options` to `categories` throughout that paragraph.

Philip-Larsen-Donnelly:
> @Philip-Larsen-Donnelly I've copied a lot of texts from "[Using the Data Entry app](https://github.com/dhis2/dhis2-docs/blob/master/src/user/using-the-data-entry-app.md)". Should I apply your suggestion to that file as well?

Understood. I don't think we need to updated the old one; that would just break the localised docs unnecessarily (because the strings change and are therefore no longer considered to be translated). I realise there is a lot of poor grammar (in my opinion :) ) across the docs, and a good time to try to catch it is when adding new content.

src/user/using-the-aggregate-data-entry-app.md
cooper-joe: Could we refer to them as `Additional selections`? `Selections` is ambiguous regarding what they actually are (options, category options, etc.) but rather refers to the action of selecting. 

I also think the wording here is a quite technical, I've attempted to make a pass at making it slightly more user friendly:

> Additional selections (dependent on previous choices)
>
> If there are any other selections available, they’ll be shown as the last controls in the top bar. Additional selections depend on the chosen data set, organisation unit, and period, so they won’t be shown until those first three selections are made. If there aren’t any additional selections, the control won’t be shown.

Mohammer5:
@Philip-Larsen-Donnelly I've addressed all change requests. Should be ready for another review now

Mohammer5:
Superseeded by #1095

PR #1089, opened 2022-10-12
Update Password and Username requirements

adityaacse:


src/user/manage-users-user-roles-and-user-groups.md
```diff
@@ -193,6 +193,20 @@ want to, you may also enter the user name that the account will have. If
 you leave the user name empty, then the user may choose their own user
 name when they respond to the invitation (as long as it is not taken
 already for another user.)<br/><br/>
+
+#### Username requirements
```
Topener: Could you please move this section below "Create account with user details" instead of the current positioning below the invite by email?

Topener (changes requested):
Only 1 change as far as I can see, I'll let someone else review as well! Thanks for the contribution

src/user/manage-users-user-roles-and-user-groups.md
adityaacse: Hi @Topener Changes made, please review.

Topener (approved):
Thanks so much for this PR, it looks all good to me!

src/user/manage-users-user-roles-and-user-groups.md
```diff
@@ -175,6 +175,18 @@ the new user such as username, password, etc. Under these conditions, the fields
...
+The following rules apply when you create a new username. The username must:
+
+  - Contain at least 4 characters.
+  
+  - Not contain more than 255 characters.
+
+  - Contain lowercase and UPPERCASE latin characters and numbers (`a-z`,`A-Z`,`0-9`).
+
+  - These characters are also allowed `. _ @` and `#`, but these may only be used as a separator, and not as a leading or trailing character, and should not be repeated (i.e. `user@@name` is not allowed).<br/><br/>
+
```
Philip-Larsen-Donnelly: Indent in line with bullet list

```suggestion

    #### Username requirements
    
    The following rules apply when you create a new username. The username must:

    - Contain at least 4 characters.
    
    - Not contain more than 255 characters.

    - Contain lowercase and UPPERCASE latin characters and numbers (`a-z`,`A-Z`,`0-9`).

    - These characters are also allowed `. _ @` and `#`, but these may only be used as a separator, and not as a leading or trailing character, and should not be repeated (i.e. `user@@name` is not allowed).

```

Philip-Larsen-Donnelly (approved):
Thanks @adityaacse!

PR #1095, opened 2022-10-17
docs: data entry beta user docs

cooper-joe:
This PR adds user documentation for the Data Entry (beta) app.

Note: this PR contains a documentation rewrite, rather than an update of the previous app documentation.

## todos:

- [x] internal linking
- [x] **Working offline** section
- [x] review _Notes_ sections from old documentation for any additional info that must be included
- [x] related links
- [x] use keys markdown_plugin for keyboard shortcuts


src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Working offline
+
+==WIP==
+
+
+## Features not supported in the beta version
+
+The **Data Entry (beta) app** is constantly evolving and new features are being added. There are some features that the previous Data Entry app offered that are supported yet:
+- **Custom forms** aren't supported yet, but we're learning more about how custom forms are used to build a sustainable, flexible solution.
```
Mohammer5: Custom forms are supported, but we don't support custom CSS and JS. So if the custom form is just a table created in the custom form editor in the maintenance app, it should work (see [here](https://debug.dhis2.org/dev/dhis-web-aggregate-data-entry/#/?attributeOptionComboSelection=LFsZ8v5v7rq-C6nZpLKjEJr_yY2bQYqNt0o-i4Nbp8S2G6A&dataSetId=lyLU2wR22tC&orgUnitId=DiszpKrYNg8&periodId=202209))!

cooper-joe: Ah, thanks for the clarification @Mohammer5. I've edited this to be clearer.

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,186 @@
...
+
+The badge in the header bar at the top of the screen shows your connection status. If you are not connected to the internet or DHIS2 server, the badge shows *offline*. As well as the offline badge, forms cells you enter data will show the *waiting to sync* status.
+
+![](resources/images/data_entry_beta/offline.png)
+
+
+## Features not supported in the beta version
+
+The **Data Entry (beta) app** is constantly evolving and new features are being added. There are some features that the previous Data Entry app offered that are supported yet:
+- **Custom forms** don't support CSS/JS yet. Custom forms that use CSS/JS won't be loaded. We're learning more about how custom forms are used to build a sustainable, flexible solution.
```
Mohammer5: I think they will be loaded but without the css and js and therefore might be broken. I'm not too sure about it, but @HendrikThePendric knows as he has implemented that part

Mohammer5 (changes requested):
Please change all `data-entry-beta-app` in the anchors to `aggregate-data-entry-app`. Then we won't have to update these when removing the beta status and circulating old links still work

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,190 @@
...
+
+![](resources/images/data_entry_beta/filter-form.png) { width=60% }
+
+To filter the whole data entry form, enter a value into the input at the top of the form.
+
+#### Filtering a section { #filtering-section-data-entry-form }
+
+![](resources/images/data_entry_beta/filter-section.png) { width=60% }
+
+For section form you can filter only inside a single section. Enter a value into the input at the top of a section.
```
Philip-Larsen-Donnelly: Sounds a little awkward/ambiguous to me. Perhaps:
```suggestion
For section forms you can also filter inside a single section. Enter a value into the input at the top of a section.
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,190 @@
...
+
+If a data value already has a comment, you can edit it by clicking the *Edit comment* button below the comment.
+
+![](resources/images/data_entry_beta/comment-edit-button.png) { width=50% }
+
+#### Minimum and maximum limits { #data-entry-value-limits }
+
+A data value can have a minimum and maximum limit. These limits won't allow someone to enter values outside of the minimum or maximum. 
+
+To add limits to a data value, [make sure the data details sidebar is open](#opening-the-data-details-sidebar), then click the *Add limits* button in the *Min and max limits* section. If possible, an average value will be shown to help you create reasonable limits.
```
Philip-Larsen-Donnelly: This should only be available to someone with the appropriate privileges, right?

cooper-joe:
> Please change all `data-entry-beta-app` in the anchors to `aggregate-data-entry-app`. Then we won't have to update these when removing the beta status and circulating old links still work

👍 Updated, using the approach from [your superseded PR](https://github.com/dhis2/dhis2-docs/pull/1088), thanks.

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,190 @@
...
+
+If a data value already has a comment, you can edit it by clicking the *Edit comment* button below the comment.
+
+![](resources/images/data_entry_beta/comment-edit-button.png) { width=50% }
+
+#### Minimum and maximum limits { #aggregate_data_entry_app.limits}
+
+A data value can have a minimum and maximum limit. These limits won't allow someone to enter values outside of the minimum or maximum. 
+
+To add limits to a data value, [make sure the data details sidebar is open](#aggregate_data_entry_app.opening_data_details_sidebar), then click the *Add limits* button in the *Min and max limits* section. If possible, an average value will be shown to help you create reasonable limits.
```
Philip-Larsen-Donnelly, line 136: This is only available for users with the correct privileges, right?

Mohammer5 (approved):

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,190 @@
...
+![](resources/images/data_entry_beta/validation-button.png) { width=60% }
+
+The validation results are shown in the details sidebar, grouped into high, medium, and low priority results. Once you've fixed validation issues, click the *Run validation again* button to recheck data values.
+
+![](resources/images/data_entry_beta/validation-results.png) 
+
+
+### Completing a form { #aggregate_data_entry_app.completing }
+
+After entering data and running validation, the last step is completing a form. Mark a form as complete by clicking the *Mark complete* button in the bottom bar.
```
Philip-Larsen-Donnelly:
```suggestion
After entering data and running validation, the last step is completing a form. This indicates that all the intended data has been entered, and that empty cells are supposed to be empty. Mark a form as complete by clicking the *Mark complete* button in the bottom bar.
```

Philip-Larsen-Donnelly (commented):
a few small suggestions

src/user/using-the-aggregate-data-entry-app.md
cooper-joe: Thanks, committed something along these lines, just changed the wording slightly.

src/user/using-the-aggregate-data-entry-app.md
cooper-joe: Yep, note added.

src/user/using-the-aggregate-data-entry-app.md
cooper-joe: Thanks, committed.

cooper-joe:
Thanks for the suggestions @Philip-Larsen-Donnelly, committed those changes.

Philip-Larsen-Donnelly (approved):
I would say this is good enough to use as the initial document at this point.

Mohammer5:
@Philip-Larsen-Donnelly You'll have to merge this branch

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,192 @@
...
+The badge in the header bar at the top of the screen shows your connection status. If you are not connected to the internet or DHIS2 server, the badge shows *offline*. As well as the offline badge, forms cells you enter data will show the *waiting to sync* status.
+
+![](resources/images/data_entry_beta/offline.png) { width=60% }
+
+
+## Features not supported in the beta version { #aggregate_data_entry_app.unsupported_features }
+
+The **Data Entry (beta) app** is constantly evolving and new features are being added. There are some features that the previous Data Entry app offered that are supported yet:
+- **Custom forms** don't support CSS/JS yet. Custom forms that use CSS/JS won't be loaded. We're learning more about how custom forms are used to build a sustainable, flexible solution.
+- **Multi-organisation unit entry** isn't supported.
```
tomzemp, line 193: I don't know if we want to advertise all of the features we do not support? With this one in particular, Lars wanted to ask people if they were still using this before deciding if we should implement?

If we are advertising non-supported features, we have also do not support validation checks for compulsory data element operands and field combinations (`fieldCombinationRequired: true` property on a data set)

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,192 @@
...
+
+If a data value already has a comment, you can edit it by clicking the *Edit comment* button below the comment.
+
+![](resources/images/data_entry_beta/comment-edit-button.png) { width=50% }
+
+#### Minimum and maximum limits { #aggregate_data_entry_app.limits}
+
+A data value can have a minimum and maximum limit. These limits won't allow someone to enter values outside of the minimum or maximum. 
+
+To add limits to a data value, [make sure the data details sidebar is open](#aggregate_data_entry_app.opening_data_details_sidebar), then click the *Add limits* button in the *Min and max limits* section. If possible, an average value will be shown to help you create reasonable limits.
```
tomzemp, line 136: Note that not all users have authority to edit/add or delete min/max values. If they do not, they will not see relevant buttons.

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,192 @@
...
+## Working offline { #aggregate_data_entry_app.working_offline }
+
+You don't need to be connected to the internet or DHIS2 server to enter data. Data entered into forms while you are offline is saved to your local computer. When you reconnect to the internet/server your locally saved data automatically syncs with your DHIS2 server.
+
+To work offline, you need to open the Data Entry (beta) app while you are connected to the internet so the data entry forms can be downloaded and stored on your local computer. Downloading of forms happens automatically in the background.
+
+The badge in the header bar at the top of the screen shows your connection status. If you are not connected to the internet or DHIS2 server, the badge shows *offline*. As well as the offline badge, forms cells you enter data will show the *waiting to sync* status.
+
+![](resources/images/data_entry_beta/offline.png) { width=60% }
+
```
tomzemp, line 182: Would it be useful to note that some* functionality is not available offline? *(Validation, History, Audits)

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,192 @@
...
+![](resources/images/data_entry_beta/validation-button.png) { width=60% }
+
+The validation results are shown in the details sidebar, grouped into high, medium, and low priority results. Once you've fixed validation issues, click the *Run validation again* button to recheck data values.
+
+![](resources/images/data_entry_beta/validation-results.png) 
+
+
+### Completing a form { #aggregate_data_entry_app.completing }
+
+After entering data and running validation, the last step is completing a form. Completing a form means that all the intended data has been entered and empty cells are intentionally left empty. Mark a form as complete by clicking the *Mark complete* button in the bottom bar.
```
tomzemp, line 86: I wonder if it would be better to say `After entering data and running validation, you can complete the form`? The current wording feels as though we are saying that a user has to complete the form whereas I think it's more of a workflow that implementers may or may not use?

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,192 @@
+# Data Entry (beta) app
+
+## About the Data Entry (beta) app { #aggregate_data_entry_app.about }
+
+The Data Entry (beta) app is used to enter aggregate data in DHIS2. Aggregate data is collected at a group level and doesn't belong to any one individual. Looking to enter individual-level data? Check out the [Tracker Capture app](#tracker_capture_app).   
+
+## What makes a data entry form? { #aggregate_data_entry_app.what_makes_a_form }
+
+Data entry forms are made from different choices, which make up the *context*:
```
tomzemp: Since we never refer to the word `context` again in the documentation, I wonder if we might not include it? It sounds a bit technical?

src/user/using-the-aggregate-data-entry-app.md
Philip-Larsen-Donnelly: I think the point here was that these (custom forms/multi OU) were explicitly dealt with in sections of the old app documentation.
I think it's fine as is for first push, but we can tweak later if necessary.

src/user/using-the-aggregate-data-entry-app.md
Philip-Larsen-Donnelly: I think that point is addressed a few lines later.

src/user/using-the-aggregate-data-entry-app.md
Philip-Larsen-Donnelly: This is a fair point. On the other hand, it probably doesn't hurt to "complete" the form, even if the implementers don't use that status in their workflow. Arguably, we should encourage its use (Completed state) instead of the "fill with zeros" approach, so I think I would stick with the current wording.

Mohammer5: > After entering data and running validation, you can complete the form

To me this wording suggests that completing a form is only possible when the form passes validation, which is not true. So to me the original wording is more workflow-oriented while the suggestion sounds like a technical description (which - if interpreted that way - would be wrong)

src/user/using-the-aggregate-data-entry-app.md
Philip-Larsen-Donnelly: True. I remember it jumped out at me a bit when I first read it. I think it would be a little softer if it wasn't in italics.
We could also have an alternative wording?

```suggestion
Data entry forms are filled in for a specific context, which is defined by the following choices:
```

cooper-joe: Yeah agreed with these points, I think @Philip-Larsen-Donnelly's suggestion is an elegant way to include it and make it not-too-technical.

src/user/using-the-aggregate-data-entry-app.md
cooper-joe: Yes, good idea. Added a note.

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,194 @@
...
+
+![](resources/images/data_entry_beta/limits-add-button.png) { width=50% }
+
+Once you've added a minimum and maximum limit, click the *Save limits* button.
+
+If a data value already has limits set up, you can change them or remove them. To change the limits, click the *Edit limits* button. To remove the limits, click the *Delete limits* button.
+
+![](resources/images/data_entry_beta/limits-edit-button.png) { width=50% }
+
+Adding, changing, or removing limits requires the correct user privileges.
```
Philip-Larsen-Donnelly: Use note/caution format and add a little info to clarify that it is not simple for the current user (not sure I'm completely happy with my text, though?)
```suggestion
> **Caution**
>
> Adding, changing, or removing limits requires the correct user privileges. When limits are applied to a data value, the same limits apply for everyone using same data set.
```

src/user/using-the-aggregate-data-entry-app.md
```diff
@@ -0,0 +1,194 @@
...
+
+You don't need to be connected to the internet or DHIS2 server to enter data. Data entered into forms while you are offline is saved to your local computer. When you reconnect to the internet/server your locally saved data automatically syncs with your DHIS2 server.
+
+To work offline, you need to open the Data Entry (beta) app while you are connected to the internet so the data entry forms can be downloaded and stored on your local computer. Downloading of forms happens automatically in the background.
+
+The badge in the header bar at the top of the screen shows your connection status. If you are not connected to the internet or DHIS2 server, the badge shows *offline*. As well as the offline badge, forms cells you enter data will show the *waiting to sync* status.
+
+![](resources/images/data_entry_beta/offline.png) { width=60% }
+
+Some features aren't available when offline because they need contact with a DHIS2 server to work. Validation, data value history, and data value audit logs aren't available offline.
```
Philip-Larsen-Donnelly: change to a note admonition
```suggestion
> **Note**
>
> Some features aren't available when offline because they need contact with a DHIS2 server to work.  
> Validation, data value history, and data value audit logs aren't available offline.
```

src/user/using-the-aggregate-data-entry-app.md
Philip-Larsen-Donnelly: In that case, let's drop that sentence
```suggestion
- **Custom forms** don't support CSS/JS yet. We are currently examining the different ways custom forms are used in order to build a more sustainable, flexible solution.
```

src/user/using-the-aggregate-data-entry-app.md
tomzemp: Okay. Sounds good then with the original language. I guess if implementers don't want some different logic for data set completion, they can take that up in their training materials

PR #1108, opened 2022-11-16
docs: edit description of scheduling events

caho1109:
#1102

### DIFF
--- a/src/user/configure-programs-in-the-maintenance-app.md
+++ b/src/user/configure-programs-in-the-maintenance-app.md
@@ -440,7 +440,7 @@ by the data capture apps to identify this program stage.
     of the stage in data entry form. This field is selected by
     default.
     5. Enter **Standard interval days**. This value will be the suggested interval between the last event in a repeatable stage and the scheduled date of the next event.
-    6. (Optional) Select a  **Default next scheduled date**. This will show a list of assigned data elements of type **date**. If an element is selected, the Tracker client will use this as the default starting date. The data element can be used by program rules to dynamically schedule intervals between events.
+    6. (Optional) Select a  **Default next scheduled date**. This will show a list of assigned data elements of type **date**. If an element is selected, the Tracker client will use this as the default scheduled date. The data element can be used by program rules to dynamically schedule intervals between events.
 4. Enter form details
 
 
--- a/src/user/configure-programs-in-the-maintenance-app.md
+++ b/src/user/configure-programs-in-the-maintenance-app.md
@@ -430,7 +430,7 @@ actions should be taken at each stage.
     2.  (Optional) select a **Color** and an **Icon** that will be used
 by the data capture apps to identify this program stage.
     3. Enter a **Description**.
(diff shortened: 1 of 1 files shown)


src/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -430,7 +430,7 @@ actions should be taken at each stage.
     2.  (Optional) select a **Color** and an **Icon** that will be used
 by the data capture apps to identify this program stage.
     3. Enter a **Description**.
-    4. In the **Scheduled days from start** field, enter the minimum
+    4. In the **Scheduled days from start** field: The first event scheduled in this  program stage will be this many days from the enrollment or the incident date - based on the configuration what date to use
 number of days to wait for starting the program stage.
```
Philip-Larsen-Donnelly: Suggest a slight change of wording here. It would be even better if we explain where the configuration is (i.e which dictates whether enrollment or incident date is used as the reference).

```suggestion
    4. Enter the required number of days into the **Scheduled days from start** field: The first event in this program stage will be scheduled this many days after the enrollment or the incident date, depending on the configuration.
```

src/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -440,7 +440,7 @@ number of days to wait for starting the program stage.
     for a repeatable stage after you click *Complete* for an event
     of the stage in data entry form. This field is selected by
     default.
-    5. Enter **Standard interval days**. The number of days to repeat the repeatable program stage.  
+    5. Enter **Standard interval days**. When scheduling an event in a repeatable stage, and there are previous events in that stage, this fields value will be the suggested interval between the latest event in the repetable stage and the scheduled date of the new event.
```
Philip-Larsen-Donnelly: I think this can be simplified a bit.
```suggestion
    5. Enter **Standard interval days**. This value will be the suggested interval between the last event in a repeatable stage and the scheduled date of the next event.
```

Bekkalizer (approved):
Looks good and precise!

src/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -440,7 +439,7 @@ number of days to wait for starting the program stage.
     for a repeatable stage after you click *Complete* for an event
     of the stage in data entry form. This field is selected by
     default.
-    5. Enter **Standard interval days**. The number of days to repeat the repeatable program stage.  
+    5. Enter **Standard interval days**. This value will be the suggested interval between the last event in a repeatable stage and the scheduled date of the next event.
     6. (Optional) Select a  **Default next scheduled date**. This will show a list of assigned data elements of type **date**. If an element is selected, the Tracker client will use this as the default starting date. The data element can be used by program rules to dynamically schedule intervals between events.
```
Bekkalizer: I know this was not touched now - but baybe we could update this sentence a bit as well? 
>Tracker client will use this as the default starting date

...default scheduled date?

caho1109: done

PR #1152, opened 2023-03-13
docs: OpenAPI endpoints documentation

jbee:
OBS! New file needs added to menu in https://github.com/dhis2/dhis2-docs-builder/pull/164


netroms (approved):

Philip-Larsen-Donnelly:
It might be useful to know what tags are available for filtering.
Is that something we could list here @jbee ?

jbee:
> Is that something we could list here

Yes, I'll add it.

PR #1153, opened 2023-03-14
fix: Update tracker docs for Characters not accepted in Tracker capture Attributes [ DHIS2-10215 ] 

lucaCambi77:
see https://github.com/dhis2/dhis2-core/pull/13285

### DIFF
--- a/src/developer/web-api/new-tracker.md
+++ b/src/developer/web-api/new-tracker.md
@@ -1342,6 +1342,11 @@ A query where multiple operand and item filters are specified for dates interval
     GET api/tracker/trackedEntities?orgUnit=DiszpKrYNg8&program=ur1Edk5Oe2n
       &filter=lw1SqmMlnfh:GT:01-01-2000 12:00:00:LT:01-01-2001 12:00:00
 
+Multiple operands do not apply to words. In that case, we can use multiple filters:
+
+    GET api/tracker/trackedEntities?orgUnit=DiszpKrYNg8&program=ur1Edk5Oe2n
+      &filter=lw1SqmMlnfh:SW:hello&filter=lw1SqmMlnfh:EW:world
+
 To query on an attribute using multiple values in an *IN* filter:
 
     GET /api/tracker/trackedEntities?orgUnit=DiszpKrYNg8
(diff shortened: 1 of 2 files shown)


src/developer/web-api/new-tracker.md
```diff
@@ -1332,13 +1332,20 @@ and one attribute is used as a filter:
...
+      &filter=lw1SqmMlnfh:GT:150:LT:190
+
+A query where multiple operand and item filters are specified for dates interval (also support milliseconds and time zone):
+
+    GET api/tracker/trackedEntities?orgUnit=DiszpKrYNg8&program=ur1Edk5Oe2n
+      &filter=lw1SqmMlnfh:GT:01-01-2000 12:00:00:LT:01-01-2001 12:00:00
+
+Multiple operands do not apply to words. In that case, we can use multiple filters:
+
+    GET api/tracker/trackedEntities?orgUnit=DiszpKrYNg8&program=ur1Edk5Oe2n
```
enricocolasante, line 1347: This request returns a validation error as it seems that filters for attributes are not allowed to be specified more than once.

enricocolasante (changes requested):
Should we change the docs, the code or I am looking at it in the wrong way?

src/developer/web-api/new-tracker.md
lucaCambi77: We have discussed it in the design and also in the PRs like https://github.com/dhis2/dhis2-core/pull/13464. This comes up only now. I didn't remember that. So range for words can't be applied then

enricocolasante: It was unexpected for me as well. I just tried the calls specified in the docs

lucaCambi77: I didn't check. What we can do is allow multiple filters for the words range in the validation. Otherwise, we need to redefine the split separator or the logic we have in place to manage all cases in the parsing

Philip-Larsen-Donnelly:
@lucaCambi77 new-tracker.md was recently changed to tracker.md, so the simplest might be to open this again against that file.

PR #1157, opened 2023-04-03
docs: users should land on new tracker by default

teleivo:
its been at least 2 years since new tracker has been released (from 2.36 according to our docs). New tracker is far from new anymore. We've added a disclaimer to the old tracker docs some time ago that anyone starting out should use the new tracker APIs. Its time to also rearrange the pages so new users land on the new tracker APIs by default.

This is how the pages should show up in the sidebar on the left

* Tracker
* Tracker (Old)

I've used the word old as its the one we are using internally and its the antonym for new. Let me know if you another name makes more sense 👌🏻


src/developer/web-api/tracker.md
```diff
@@ -1,531 +1,1357 @@
...
->* If you are using the tracker endpoints in your implementation, it is
->  advisable to use the new version at [New Tracker - DHIS2 Documentation](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/new-tracker.html)
->* If you already use the tracker endpoints in production, please plan to
->  migrate over to the new version. Reach out on the [community of practice](https://community.dhis2.org)
->  if you need any assistance. NOTE: The feature for data sync(importMode=SYNC)
->  is not implemented in the new tracker endpoints, and if you are using this
->  feature you will have to postpone the migration until a new SYNC feature is
->  in place.
+> - The old endpoints are marked as deprecated but still work as before.
+> - Some functionality is not yet ready in the new endpoints, but they support their primary use-cases.
```
enricocolasante: I would rephrase that and maybe specify that just sync is not working in the new endpoints but everything else is ready

enricocolasante (approved):

PR #1183, opened 2023-05-16
docs: Added docs for custom icon feature [DHIS2-15039]

muilpp:


src/developer/web-api/metadata.md
```diff
@@ -1235,17 +1237,74 @@ actual icon.
...
+  fileResourceUid: "ohUVXsOZ8qp",
+  userUid: "AIK2aQOJIbj",
+  href: "<dhis server>/api/fileResources/ohUVXsOZ8qp/data"
+}
+```
 
-The keywords can be used to filter which icons to return. Passing a list
-of keywords with the request will only return icons that match all the
-keywords:
+It's also possible to get a particular icon direcly by filtering by the its key, in the example below, the key is mosquito_outline.
```
enricocolasante, line 1256:
```suggestion
It's also possible to get a particular icon directly by filtering by its key, in the example below, the key is mosquito_outline.
```

stian-sandvold (approved):

PR #1191, opened 2023-06-19
docs: maps download and value labels

turban:
Update to the DHIS 2 maps user guide about the new download mode and value labels introduced in version 2.40.


turban:
Moved to: https://github.com/dhis2/maps-app/pull/2691/

PR #1200, opened 2023-06-28
chore: program copy editorial changes

Philip-Larsen-Donnelly:
@david-mackessy 

I think your text for this documentation is great. I did't manage to review before the merge, so I just wanted to push these changes to share a couple of editorial points:
1. I always try to encourage people to put blank lines between all markdown blocks, including headers, etc. It sounds a bit unnecessary, but different markdown parsers detect blocks in slightly different ways, and keeping the blank lines avoids some problems I have seen.
2. We have a special way to treat "notes" and other admonitions (https://docs.dhis2.org/en/implement/support-and-documentation/dhis2-documentation-guide.html#admonitions)

Finally a quick question about the actual implementation.
Why do we only offer a "prefix" for the new program name, and not the option to provide a complete new name?

Thanks!

### DIFF
--- a/src/developer/web-api/metadata.md
+++ b/src/developer/web-api/metadata.md
@@ -3267,10 +3267,14 @@ The `fields` parameter can be used to narrow the fields included for the shown o
 | MULTI_TEXT | None
 
 ## Copy Program
+
 ### Introduction
+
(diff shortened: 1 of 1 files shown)


david-mackessy (approved):

david-mackessy:
Thanks for the feedback, I'll take those on for future PRs.

> Finally a quick question about the actual implementation. Why do we only offer a "prefix" for the new program name, and not the option to provide a complete new name?

We could offer this as an option, it never actually came up in the requirements or other conversations. That would be another request parameter which should be a small change if implemented.

PR #1207, opened 2023-07-14
docs: Add 'ieq' operator for case insensitive string filter [DHIS2-12167]

david-mackessy:
Add 'ieq' operator for case insensitive string filter
- add example of usage

### DIFF
--- a/src/developer/web-api/metadata.md
+++ b/src/developer/web-api/metadata.md
@@ -286,35 +286,35 @@ objects, is supported as well.
 
 Table: Available Operators
 
-| Operator | Types                                                                                                         | Value required | Description                                     |
-|----------|---------------------------------------------------------------------------------------------------------------|----------------|-------------------------------------------------|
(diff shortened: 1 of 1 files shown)


src/developer/web-api/metadata.md
```diff
@@ -323,6 +324,10 @@ Get data elements with id property ID1 or ID2:
 
     /api/dataElements?filter=id:eq:ID1&filter=id:eq:ID2
 
+Get data elements, ignoring case, with name property Data Element One:
+
+    /api/dataElements?filter=name:ieq:data element one
```
jbee: I think spaces in an URL is not a good example as they are not allow unnecessarily complicating the example with a problem that is not about what this wants to demonstrate.

david-mackessy: updated

larshelge (approved):

PR #1211, opened 2023-07-20
feat: add Data Element and webhook as "who to send it?" notification

dayures:


dayures:
@Wonfrost maybe this could be improved.

PR #1218, opened 2023-08-03
docs: Update data set notification template [DHIS2-13380]

david-mackessy:
while working on DHIS2-13380 I noticed that these docs were out of date

### DIFF
--- a/src/developer/web-api/metadata.md
+++ b/src/developer/web-api/metadata.md
@@ -1868,19 +1868,24 @@ JSON payload sample is given below:
 ```json
 {
   "name": "dataSetNotificationTemplate1",
-  "notificationTrigger": "COMPLETION",
+  "dataSetNotificationTrigger": "DATA_SET_COMPLETION",
   "relativeScheduledDays": 0,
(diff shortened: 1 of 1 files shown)


src/developer/web-api/metadata.md
```diff
@@ -1868,20 +1868,25 @@ JSON payload sample is given below:
 ```json
 {
   "name": "dataSetNotificationTemplate1",
-  "notificationTrigger": "COMPLETION",
+  "dataSetNotificationTrigger": "DATA_SET_COMPLETION",
```
jbee, line 1871: Just curious, was this using a incorrect constant name before or are these different things or different names for the same thing?

jbee (approved):

src/developer/web-api/metadata.md
david-mackessy: looks like the field name and constant value were renamed back in 2017/2018 (revision numbers: 8ae510b244cad8c8345724b23593d1dbb82f71ab & 
dbc89a2073eb0a463a71853f4ea85e6b4abd67fc)
The example didn't work for me but updating to these new values worked.

PR #1240, opened 2023-10-09
chore: removed duplicated authority

dayures:
Keep the text of the one that appears in the response of the API https://play.dhis2.org/40.1.0/api/authorities


jason-p-pickering (approved):

jason-p-pickering (changes requested):
Looks like there are some merge conflicts. Could you have another look?

PR #1250, opened 2023-10-27
docs: Date and time parameters in tracker [DHIS2-16019]

enricocolasante:


src/developer/web-api/tracker.md
```diff
@@ -1261,20 +1261,20 @@ The endpoint returns a list of tracked entities that match the request parameter
 |`programStatus`|`String`|`ACTIVE`&#124;`COMPLETED`&#124;`CANCELLED`|The ProgramStatus of the Tracked Entity Instance in the given program|
 |`programStage`|`String`|`UID`|a Program Stage `UID` for which instances in the response must have events for|
 |`followUp`|`Boolean`|`true`&#124;`false`|Indicates whether the Tracked Entity Instance is marked for follow up for the specified Program|
-|`updatedAfter`|`DateTime`| [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) | Start date for last updated|
-|`updatedBefore`|`DateTime`| [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) | End date for last updated|
+|`updatedAfter`|`DateTime`| [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) | Start date and time for last updated|
```
muilpp, line 1264: reading the description of this one and the one below, it's not clear to me what do they return.
last updated means the last time it was updated, right? why do we need a start date and an end date?

enricocolasante: This is export endpoint for Tracked Entities, the client is filtering by last updated. The client would never want to filter for an exact match on the last updated but wants to filter starting from a date, or ending to a date or between dates.
Or in other terms, before and after last updated

muilpp (approved):

PR #1268, opened 2023-12-13
fix: Add ref. to tracker OU description

maikelarabori:
Add URL reference to OU Tracker documentation/description.


muilpp:
Looks good to me. The only thing to mention is that there are three other modes not stated in the docs, they are `ACCESSIBLE`, `CAPTURE` and `ALL`. Are they not used in analytics?

muilpp (approved):

PR #1272, opened 2023-12-19
docs: Document available ordering properties in 2.40

muilpp:


enricocolasante (approved):

teleivo:
@muilpp I think this change is incorrect 🤔 I think you added the changes to `Common request parameters` suggesting users can `order` by these properties on all export endpoints. I think the only property that is supported across endpoints is `createdAt`. Please revert/put the change to the endpoint you wanted to describe 😄 

You can take https://github.com/dhis2/dhis2-docs/blob/03237d0ddda760b25635eaaaecd7bb5afc9e8b4c/src/developer/web-api/tracker.md?plain=1#L1600 as a blueprint. I don't think java classes like the OrderCriteria belong in any API doc.

PR #1276, opened 2024-01-02
Update "hide program stage" documentation

karolinelien:
Update "Hide program stage" name to "Prevent adding new events to stage" and description


src/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -1193,7 +1193,7 @@ objects:
...
         | **Hide field** | **Data element to hide**<br>         <br>**Tracked entity attribute to hide**<br>         <br>**Custom message for blanked field** | Used when you want to hide a field from the user.<br>         <br>**Custom message for blanked field** allows you to define a custom message displayed to the user in case the program rule hides and blanks out the field after the user typed in or selected a value.<br>         <br>If a hide field action hides a field that contains a value, the field will always removed. If no message is defined, a standard message will be displayed to alert the user. |
         | **Hide section** | **Program stage section to hide** | Used when you want to hide a section in a program stage from the user.  |
-        | **Prevent adding new events to stage** | **Program stage where new events should be prevented** | Used when you want to prevent new events to be added to a program stage. Existing events will not be hidden.
+        | **Prevent adding new events to stage** | **Program stage where new events should be prevented** | Used when you do not want users to add any more events to a program stage. Existing events will not be hidden.
```
linadhis2: Would rename "Program stage where new events should be prevented" as it's the same as the title itself. 

Maybe into: "Restrict users to adding new events to the program stage" 

Otherwise LGTM

karolinelien: "Program stage where new events should be prevented" is explaining what you are selecting in the dropdown specifically, so should say "Program stage [action of program rule]". So could be "Program stage where users will not be able to add new events"?

linadhis2: Ahh, thank you for explanation! Yes, perfect!

cooper-joe: I think `Program stage where users will not be able to add new events` is OK, there aren't many other ways to word this in the current pattern we're using.

linadhis2 (approved):

linadhis2 (approved):
LGTM

Birkbjo (approved):

PR #1296, opened 2024-01-16
add "including zero values in analytics tables"

karolinelien:


Philip-Larsen-Donnelly:
Maybe we should try to clarify things here a little? i.e. is it correct to say that if this setting is off, then zero values will NOT be stored, regardless of the "store zero data values" property on the DEs?

jason-p-pickering:
This property only controls whether zeros will be written to the analytics tables. "Store zero data values" refers to whether zeros will be stored or ignored when data is actually entered.

Philip-Larsen-Donnelly (approved):

PR #1302, opened 2024-01-25
[docs] Remove "Put analytics into maintenance mode"

linadhis2:
As a part of https://dhis2.atlassian.net/browse/DHIS2-16534, updating the documentation on master and 40 where this setting is no longer applicable.


tomzemp:
@linadhis2: Thanks! Do you also want to review it from the developer docs with this PR: https://github.com/dhis2/dhis2-docs/blob/master/src/developer/web-api/settings-and-configuration.md?plain=1#L153 ?

(If you want to do that separately, I can approve, but it seems it will auto-merge if I just give the approval in advance)

linadhis2:
Thanks @tomzemp - good catch, I saw it but wasn't sure. I've added an update to that file in both commits, too!

karolinelien (approved):
LGTM!

PR #1319, opened 2024-02-12
DHIS2-16396: document db.pool.type config option

cjmamo:


david-mackessy:
It would be nice to somehow transfer the good information in the Jira issue into the docs somehow (why unpooled might be preferable). This info will not be easily available to a user reading the docs.

david-mackessy (approved):

PR #1337, opened 2024-02-26
docs: JDK 17 for v41

david-mackessy:
converted sentences to compatibility matrix

### DIFF
--- a/src/sysadmin/installation.md
+++ b/src/sysadmin/installation.md
@@ -52,10 +52,19 @@ Later DHIS2 versions require the following software versions to operate.
 
 - An operating system for which a Java JDK or JRE version 17 exists. Linux is recommended.
 - Java JDK. OpenJDK is recommended.  
-    - For DHIS 2 version 2.41, JDK 17 is required.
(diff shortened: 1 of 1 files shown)


netroms (approved):

src/sysadmin/installation.md
```diff
@@ -327,16 +328,10 @@ sudo systemctl restart postgresql
 
 ### Java installation { #install_java_installation } 
 
-The recommended Java JDK for DHIS 2 is OpenJDK 11 (for version 2.35 and later). You can install it with the following command:
+The recommended Java JDK for DHIS 2 is OpenJDK 17 (for version 2.41). You can install it with the following command:
```
larshelge: I would actually say "for version 2.40 and later". JDK 17 is well tested and performs a bit better than 11 on 2.40. This way it is also future proof and will be correct in a year or two, we will definitely forget to update this again :)

david-mackessy: lovely 👍, i'll update

netroms (approved):

PR #1338, opened 2024-02-29
fix: tracker doc clean

lucaCambi77:
- use comment instead of <tag> in JSON
- update ou modes link

### DIFF
--- a/src/developer/web-api/tracker.md
+++ b/src/developer/web-api/tracker.md
@@ -217,11 +217,9 @@ Tracker notes do not have their dedicated endpoint; they are exchanged as part o
 ```json
 {
   "trackedEntity": "oi3PMIGYJH8",
-  "_comment" : "tracked entity details ...",
   "enrollments": [
     {
       "enrollment": "EbRsJr8LSSO",
-      "_comment" : "enrolmment details ...",
       "notes": [
         {
           "note": "vxmCvYcPdaW",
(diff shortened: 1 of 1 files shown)


src/developer/web-api/tracker.md
```diff
@@ -217,38 +217,36 @@ Tracker notes do not have their dedicated endpoint; they are exchanged as part o
 ```json
 {
   "trackedEntity": "oi3PMIGYJH8",
-  <entity_details>,
-  ],
+  "_comment" : "tracked entity details ...",
```
teleivo: Why do you add `_comment`?

lucaCambi77: It Is the only way to add comments (I mean it could also be "comment") in JSON I see and it is better than the current solution. It is just an idea if we can come up with something different it is ok for me

muilpp: if "comment" is free to use, why do we add the underscore?

muilpp (approved):

src/developer/web-api/tracker.md
teleivo: why do we even need this comment? `"tracked entity details ..."` isn't all that useful. Adding the field can mislead someone to believe its part of the payload. I would remove it.

Also this will be removed once we publish the OpenAPI spec to which we have added descriptions.

PR #1351, opened 2024-03-14
docs: add docs for scheduler queues

ghost:
- https://docs.dhis2.org/en/use/user-guides/dhis-core-version-240/maintaining-the-system/scheduling.html
- https://dhis2-scheduler.netlify.app/#/

In the second commit I ran prettier to wrap the text at 80 columns, because the original doc was a bit messy. I did not make changes to the text in the second commit, other than the formatting changes that prettier made. If you want a simpler diff see the first commit, where the actual changes to the text were made.

## Images

I've updated the images so that they match the current design. I kept all details as similar as possible. The sizes are a little different because the old images were larger than what the instructions in this repo's readme dictate.

The new images were taken with a viewport of 1440 x 900, but since the existing images in the docs were a little cropped and I matched those, they ended up a little smaller.

If we want images that are 1440 x 900, and don't mind that they'll look a little different, we could use these:

delete_job
new_job
new_queue
overview


src/user/scheduling.md
```diff
@@ -418,3 +424,55 @@ order of their names (comparing Unicode character values).
 
 If both individual predictors and predictor groups are selected in the same
 job, the individual predictors run first, followed by the predictor groups.
+
+## Creating a queue { #scheduling_create_queue } 
+
+1.  Open the **Scheduler** app and click the "New queue" button in the top
+    right corner.
+
+1.  Choose a suitable **Name** for the new queue.
```
jbee, line 472: Does this render correctly with using `1.` multiple times?

src/user/scheduling.md
```diff
@@ -418,3 +424,55 @@ order of their names (comparing Unicode character values).
...
+switches in the **On/off** column on the landing page of the Scheduler app.
+
+Further editing of queues:
+
+1.  Click the "Actions" button of the queue you want to edit and click "Edit".
+
+1.  When done editing, press the **Save** button to persist the changes.
+
+1. If jobs were removed from the queue, they will be displayed again in the overview. But since they were part of a queue, they
+will be disabled and without a schedule.
```
jbee: This wording is a bit confusing. It seems to say the same thing twice. I think we can simplify it (same again below for deleting a queue)

src/user/scheduling.md
ghost: Yeah it does, it's a convenient way to allow inserting items in a list easily, without having to renumber the entire list.

src/user/scheduling.md
ghost: So what I'm attempting to communicate here is:

- When removing a job from a queue, it:
  - Will be displayed as a regular job again
  - Will be disabled
  - Will have its (cron or delay) schedule cleared

I don't see the repeating of the same info you're referring to. Maybe the wording makes it seem as if I'm saying the same thing twice? Or do you mean that there's repetition between this bit, and then the identical text below for deleting a queue?

kabaros (approved):
looks very good and very comprehensive to me 👌🏿

ghost:
> looks very good and very comprehensive to me 👌🏿

Thanks @kabaros. Since I've not received any other feedback I'll merge this. We can improve it further if need be.

PR #1357, opened 2024-03-22
fix: updated tracker docs for csv and compression

lucaCambi77:
Clean up and update docs for CSV and compression support.

### Approach

The docs first describe the different tracker objects and their fields. Then, we show some examples of imports and exports.
CSV definition follows the tracker objects fields, although with some exceptions.
This PR's approach is to show all the fields in the export part to document our use of CSV. In the CSV import part, currently only for Events, we link to the export section for more details about the fields. 

### Work done

- Rename Tracker Notes to Notes and make all tracker objects in the plural when headings
- add missing potentialDuplicate property for Tracked Entity
- Rename the CSV Event Payload to a generic CSV import. Currently, we only support Events CSV  import, but we make this generic so we can add to this section for others to come
- CSV and Compression support are now as an intro to the export section and for every Tracker object that supports them
- CSV export and import examples are in a raw format. I haven't found a better way to show it. Using a table might make it a better look, but it doesn't allow the user to copy and paste the examples
- CSV field definition added, and response content type with JSON and CSV are shown where relevant


src/developer/web-api/tracker.md
```diff
@@ -187,7 +188,7 @@ of the relationship must conform to are based on the `Relationship Type` of the
 >}
 >```
```
teleivo: There might be a whitespace issue here 🤔 I have seen this before. Not sure if this would render properly, have you tried to preview https://github.com/dhis2/dhis2-docs-builder

It might be necessary to add a space after every `>` so `> `

src/developer/web-api/tracker.md
```diff
@@ -187,7 +188,7 @@ of the relationship must conform to are based on the `Relationship Type` of the
...
 
-### Attribute
+### Attributes
 
 `Attributes` are the actual values describing the `tracked entities`. They can either be connected
```
teleivo, line 193: I would not quote words unless you mean a field in a JSON. I checked and see that you did not write this. It probably shows up in the diff because of the whitespace issue described above. You can remove the quotes.

src/developer/web-api/tracker.md
```diff
@@ -637,21 +638,25 @@ section](#webapi_nti_import_summary).
...
-### CSV Events payload
+### CSV import
 
-In order to maintain compatibility with older versions of tracker, the API allows to import events
-using the CSV format. As this format does not allow list as field, every row of the CSV payload
-represents an event and a data value. So for events with multiple data values, the CSV file will
-have `x` rows per event where `x` is the number of data values in that event. Other fields that are
-lists as ***relationships*** and ***notes*** are not supported. To import a CSV payload, the content
-type of the request must be set to ***application/csv*** or ***text/csv***.
+The API also allows imports based on a CSV definition. You can make a `POST` request with the body as the CSV input file. The request's `Content-Type` must be set to ***application/csv*** or ***text/csv***.
```
teleivo: We can probably shorten these sentences

To import events using CSV make a `POST` request with the CSV body and `Content-Type` set to ***application/csv*** or ***text/csv***.

src/developer/web-api/tracker.md
```diff
@@ -637,21 +638,25 @@ section](#webapi_nti_import_summary).
...
 
-#### ***CSV PAYLOAD*** example
+#### Events
 
-|event|status|program|programStage|enrollment|orgUnit|occurredAt|scheduledAt|dataElement|value|storedBy|providedElsewhere
-|---|---|---|---|---|---|---|---|---|---|---|---|
-|V1CerIi3sdL|COMPLETED|IpHINAT79UW|A03MvHHogjR|CCBLMntFuzb|DiszpKrYNg8|2020-02-26T23:00:00Z|2020-02-27T23:00:00Z|a3kGcGDCuk6|11|admin|false
-|V1CerIi3sdL|COMPLETED|IpHINAT79UW|A03MvHHogjR|CCBLMntFuzb|DiszpKrYNg8|2020-02-26T23:00:00Z|2020-02-27T23:00:00Z|mB2QHw1tU96|[-11.566044,9.477801]|admin|false
+Every row of the CSV payload represents an event and a data value. So, for events with multiple data values, the CSV file will
+have `x` rows per event, where `x` is the number of data values in that event. Other fields listed as ***relationships*** and ***notes*** are not supported.
```
teleivo: > Other fields listed as ***relationships*** and ***notes*** are not supported.

What does that mean?

src/developer/web-api/tracker.md
```diff
@@ -1432,8 +1437,11 @@ Tracker export endpoints allow you to retrieve the previously imported objects w
 
 > **NOTE**
 >
-> - All these endpoints currently support `JSON`. `CSV` is only supported by tracked entities and
->   events.
+> - All these endpoints default to a `JSON` response content. `CSV` is only supported by tracked entities and events. 
+> - You can export a CSV file by adding the `Accept` header ***text/csv*** or ***application/csv*** to the request. 
+> - Compression support is also available in zip and gzip formats for some endpoints.
```
teleivo: For which ones?

You mention it for csv so if possible it would be good to mention which ones here too.

> `CSV` is only supported by tracked entities and events.

src/developer/web-api/tracker.md
```diff
@@ -1492,6 +1500,48 @@ Two endpoints are dedicated to tracked entities:
...
+  - value (String)
+  - storedBy (String)
+  - createdBy (Username of user)
+  - updatedBy (Username of user)
+
+See [Tracked Entities](#tracked-entities) and [Attributes](#attributes) for more field descriptions.
+
+#### GZIP
+
+The response is a default `trackedEntities.csv.gz` file containing the `trackedEntities.csv` file.
```
teleivo: > is a default `trackedEntities.csv.gz` file

reads a bit odd (same for zip) 🤔 is it 

The response is file `trackedEntities.csv.gz` containing the `trackedEntities.csv` file.

src/developer/web-api/tracker.md
```diff
@@ -1492,6 +1500,48 @@ Two endpoints are dedicated to tracked entities:
 - `GET /api/tracker/trackedEntities/{id}`
   - retrieves a tracked entity given the provided id
 
+#### Response content types
+
+If not otherwise specified, JSON is the default response for the `GET` method.
```
teleivo, line 1520: Can you keep this as a paragraph. We add so many newlines in our docs that are odd.

teleivo (changes requested):
Please configure max width of 100 chars as we do for Java. Every paragraph is now formatted like this in here except for tables.

And please undo any changes to the JSON responses which I just fixed 😬

src/developer/web-api/tracker.md
lucaCambi77: This was already there, I guess it means in CSV, we can't export nested properties

src/developer/web-api/tracker.md
lucaCambi77: The ones just mentioned. We might want to say all the tracker export endpoints

src/developer/web-api/tracker.md
lucaCambi77: I haven't done that, and there are many cases already, so if we want to do it, we can do it in a separate PR. This looks out of the scope for me here

src/developer/web-api/tracker.md
lucaCambi77: I am not sure what happened, i haven't touched that part so it looks like something weird with the formatting I need to check

lucaCambi77:
> And please undo any changes to the JSON responses which I just fixed 😬

I need to add a response with multiple records otherwise it looks weird we get only 1 record from the collection endpoint. Also, that is the actual response I have from the server so I don't the reason why it is not good @teleivo

src/developer/web-api/tracker.md
teleivo: In

> Compression support is also available in zip and gzip formats for some endpoints. 

`some` is raising the question which ones. I cannot see from this diff if there would be some indentation in the bullets so it is clear which ones. Otherwise say which ones instead of `some` please.

teleivo:
> > And please undo any changes to the JSON responses which I just fixed 😬
> 
> I need to add a response with multiple records otherwise it looks weird we get only 1 record from the collection endpoint. Also, that is the actual response I have from the server so I don't the reason why it is not good @teleivo

I just replaced it and it is a response that you can get with pageSize=1. This way the JSON is valid and shorter. People can use their imagination with adding another element in the JSON array I would say.

src/developer/web-api/tracker.md
lucaCambi77: ah ok i thought you mentioned `All these endpoints`

src/developer/web-api/tracker.md
lucaCambi77: I have issues installing the docs-builder, so I have fixed it by adding a space as you suggested. Also, I've checked if there were others. However, I have tried with many live previews and couldn't see any issue

src/developer/web-api/tracker.md
teleivo: Can you please remove that sentence as it's confusing. You already state that the `fields` query parameter has no effect and that this endpoint always returns the listed fields.

src/developer/web-api/tracker.md
teleivo: You don't seem to have added what I meant.

```
>}
>```
```

should be

```
> }
> ```
```

lucaCambi77: yes i didn't push yet

src/developer/web-api/tracker.md
```diff
@@ -1536,20 +1545,18 @@ See [Tracked Entities](#tracked-entities) and [Attributes](#attributes) for more
 
 #### GZIP
 
-The response is a default `trackedEntities.csv.gz` file containing the `trackedEntities.csv` file.
+The response is `trackedEntities.csv.gz` file containing the `trackedEntities.csv` file.
```
```suggestion
The response is file `trackedEntities.csv.gz` containing the `trackedEntities.csv` file.
```

src/developer/web-api/tracker.md
```diff
@@ -1986,6 +2065,59 @@ Two endpoints are dedicated to events:
 - `GET /api/tracker/events/{id}`
     - retrieves an event given the provided id
 
+If not otherwise specified, JSON is the default response for the `GET` method. 
+The API also supports CSV export for single and collection endpoints. Furthermore, it supports 
+compressed JSON and CSV for the collection endpoint.
```
teleivo, line 2079: There seem to be newlines in your sentences that I don't get like 

Furthermore, it supports 
compressed JSON and CSV for the collection endpoint. 

Why is there one after 'it supports'?

teleivo (approved):

src/developer/web-api/tracker.md
lucaCambi77: ok sorry i didn't put the automatic return after 100

PR #1359, opened 2024-03-26
clean: tracker docs

lucaCambi77:
Following https://github.com/dhis2/dhis2-docs/pull/1357

- try to fix the [nested list](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker.html#webapi_nti_export) inside a blockquote that is not showing correctly. Not sure why `-` is in use for some lists. Markdown syntax [guide](https://confluence.atlassian.com/bitbucketserver/markdown-syntax-guide-776639995.html#Markdownsyntaxguide-Unorderedlist) would suggest `*`. Not sure also if this is an error that occurs when the docs are built. Currently, I am not able to use the https://github.com/dhis2/dhis2-docs-builder
- Fix misplaced `file` word

### DIFF
--- a/src/developer/web-api/tracker.md
+++ b/src/developer/web-api/tracker.md
@@ -1448,16 +1448,16 @@ Tracker export endpoints allow you to retrieve the previously imported objects w
 
 > **NOTE**
 >
-> - All tracker export endpoints default to a `JSON` response content. `CSV` is only supported 
+> * All tracker export endpoints default to a `JSON` response content. `CSV` is only supported 
 >   by tracked entities and events.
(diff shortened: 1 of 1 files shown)


teleivo (approved):
`-` or `*` are valid to render a list https://commonmark.org/help/

You can try https://spec.commonmark.org/dingus/
At least two spaces are needed for it to be rendered as a sublist. Not sure if the `Note` block has some influence here. That is maybe rendered by some plugin as I don't think it's part of the commonmark spec.

lucaCambi77:
@teleivo previews seem to work all good. So I guess it is something about the block that is not working

PR #1367, opened 2024-04-26
docs: Adding section on how sharing is used with aggregate data exchanges

jason-p-pickering:


src/developer/web-api/data-exchange.md
```diff
@@ -39,6 +39,20 @@ It is recommended to either specify basic authentication or PAT authentication.
...
+taken to ensure that only authorized users have access to actually submit data which results from the exchange.
+
+The following table summarizes how sharing can be used with aggregate data exchanges.
+
+
+| Sharing | Effective permissions                                                              |
+| -------- |-----------------------------------------------------------------------------------|
+| "r-------" | Can view metadata of the data exchange. |
+| "-w------" | Can edit metadata of the data exchange. |
+| "--r-----" | Can view data which is part of the exchange. |
```
vietnguyen, line 53: Correct me if i'm wrong. Data Read is not implemented in AggregateDataExchange level. Instead, it is controlled by sharing on the individual metadata objects (data elements, indicators) which are part of the request itself. The sharing on the individual “dx” objects control whether the user has access to the data or not.

jason-p-pickering:
That is not my understanding. The sharing attribute '--r-----' should mean
that the user can see the results of the exchange. You are correct that
data element and indicator sharing will also affect what the user is able
to see within the request, but the intent was that only certain users
should be able to see the exchange. I am pretty sure this is how it works
but maybe it was not implemented in this manner?

vietnguyen (approved):

PR #1376, opened 2024-05-16
fix: remove custom js css mentions, use US spelling

tomzemp:
This PR:
- removes the mentions of custom js and css under the "Login App customization" header. These still exist in the system, but were primarily used for login page customization at this point.
- updates to use American spellings (customization, rather than customisation) in line with the guidance in our design principles.

### DIFF
--- a/src/developer/web-api/settings-and-configuration.md
+++ b/src/developer/web-api/settings-and-configuration.md
@@ -527,11 +527,11 @@ DHIS2 web application. This can be particularly useful in certain situations:
   - Including CSS styles which are used in custom data entry forms and
     HTML-based reports.
 
-## Login App customisation { #login_app_customisation }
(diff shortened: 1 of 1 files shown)


Birkbjo (approved):

src/developer/web-api/settings-and-configuration.md
```diff
@@ -527,11 +527,11 @@ DHIS2 web application. This can be particularly useful in certain situations:
   - Including CSS styles which are used in custom data entry forms and
     HTML-based reports.
 
-## Login App customisation { #login_app_customisation }
+## Login App customization { #login_app_customization }
```
Philip-Larsen-Donnelly, line 530: I would generally avoid changing an existing anchor just for typo/spelling (as it may break existing links) - but this doesn't seem so critical.

PR #1386, opened 2024-05-30
chore: update tracker docs DHIS2-17064

teleivo:
* shrink the number of top level headings. The table of content is shown up to level 3 in the docs sidebar. It is very noisy right now. We have lots of these headings which can be merged. For example `SYNC and ASYNC` relates to a request parameter so it can be moved into that level 3 heading.
* add level 3 heading for create, update, delete as this is definitely something a user would want to do and find straight away.
* describe how to update/delete attribute and data values
* describe how to delete

```
## Tracker Import (`POST /api/tracker`)
### Request parameters
### Payload
### Create
### Update
### Delete
### CSV import
```

We can try to figure out if we can shrink/move the `### Import Summary` related sections which are too prominent for my taste 😁 but we should keep in mind that a lot of the docs can be removed when we publish the OpenAPI spec 😅 

* fix some broken links and make our references more stable using https://github.com/dhis2/dhis2-docs/tree/39f2b02429ab1ffc225b721051773ef12e2395a4?tab=readme-ov-file#section-references

## Next PRs
* rename any `tei` in the relative links and try to remove the old tracker docs directing to the migration guide for v41
* double check all the properties in the new tracker doc in Tracker Objects

We will then split up subtasks to add missing sections to the docs like potential duplicates, working lists, program notifications, ...


src/developer/web-api/tracker.md
```diff
@@ -550,6 +556,21 @@ Examples for the **FLAT** and the **NESTED** versions of the payload are listed
...
+is that the client does not need to provide UIDs for these references, as this is handled
+automatically.
+
+> **NOTE**
+>
+> Although nested payloads can be easier for clients to manage, the payload will always be flattened
+> before the import. For large imports, using a flat structured payload offers more control and
+> reduces overhead during the import process.
+> 
+> That being said, you cannot use nested payloads to create relationships. This is because it is not
```
enricocolasante: `you cannot use nested payloads to create relationships`.
We should be more clear here, we can create relationships in nested payloads, like a payload with 2 TEs and one relationship nested in a TE will work.
What is not possible is to have new entities embedded in the relationship, hence relationships are always leaves of the payload

src/developer/web-api/tracker.md
```diff
@@ -611,32 +632,191 @@ Examples for the **FLAT** and the **NESTED** versions of the payload are listed
...
+The payload must include all fields of the object you are updating, even if they have not been
+modified. The only exception is collections. Items in a collection that should not be changed can be
+omitted, as demonstrated in [update attribute values](#update-data-values) and [update data
+values](#update-data-values).
 
-Examples of the **ASYNC** response is shown below. For **SYNC** response, look at the [importSummary
-section](#webapi_nti_import_summary).
+Updating is not permitted in the following cases:
+* when the tracker object has already been deleted.
+* when dealing with a relationship.
```
enricocolasante: I am wondering if this is clear enough or it sounds like for example if you are updating an enrollment with a relationship then it is not possible.

src/developer/web-api/tracker.md
```diff
@@ -611,32 +632,191 @@ Examples for the **FLAT** and the **NESTED** versions of the payload are listed
...
+          "valueType": "TEXT",
+          "value": "Johnny"
+        }
+      ]
+    }
+  ]
+}
+```
+
+Note that it is not necessary to specify the tracked entity's enrollments and events. However, you
```
enricocolasante:
```suggestion
Note that it is not necessary to specify the tracked entity's enrollments, events and relationships. However, you
```
Should we even say events here, they are not really a tracked entity's field

src/developer/web-api/tracker.md
```diff
@@ -611,32 +632,191 @@ Examples for the **FLAT** and the **NESTED** versions of the payload are listed
...
+    },
+    {
+      "trackedEntity": "Gjaiu3ea38E",
+    }
+  ]
+}
+```
+
+All the children of a tracker object will be deleted if the user making the request has the
+`DELETE_CASCADE` authority.
```
enricocolasante: The 2 authorities needed here are `F_TEI_CASCADE_DELETE` and `F_ENROLLMENT_CASCADE_DELETE`

enricocolasante (approved):

PR #1394, opened 2024-06-05
docs: Adapt Program Notification Template to new API [DHIS2-17502]

muilpp:


src/developer/web-api/tracker.md
```diff
@@ -2861,3 +2861,64 @@ See an example payload below:
 }
 ```
 
+## Program Notification Template
+
+The Program Notification Template feature enables you to create message templates that can be triggered by various types of events. These templates, including both the message and subject, will be dynamically translated into actual values and sent to a configured destination. Depending on whether the notification recipient is external or internal, each template will be transformed into either a `MessageConversation` object or a `ProgramMessage` object. These intermediate objects will only contain the translated message and subject text.
```
teleivo, line 2866: Why are the intermediate objects? Is that important for users to know?

src/developer/web-api/tracker.md
```diff
@@ -2861,3 +2861,64 @@ See an example payload below:
...
+	"notificationRecipient": "ORGANISATION_UNIT_CONTACT",
+	"notifyParentOrganisationUnitOnly": false,
+	"displayMessageTemplate": "Case notification A{h5FuguPFF2j}",
+	"messageTemplate": "Case notification A{h5FuguPFF2j}",
+	"deliveryChannels": [
+		"EMAIL"
+	]
+}
+
+```
```
teleivo, line 2889: newline

src/developer/web-api/tracker.md
teleivo, line 2866: Are these templates metadata?

If so we would not need to add `Retrieving and deleting Program Notification Template` apart from a parameters table for the ones we add on top of the CrudController. It would be enough to add something like

> Since working lists are metadata refer to [metadata](#webapi_metadata) on how to create, update and delete metadata.

ameenhere:
Closing as there seems to be another PR that was merged with the same issue number. https://github.com/dhis2/dhis2-docs/pull/1425

PR #1410, opened 2024-06-28
docs: Update tracked entities ownership [DHIS2-17236]

muilpp:
Update the ACL documentation with the last changes we did on [DHIS2-17236](https://dhis2.atlassian.net/browse/DHIS2-17236)

[DHIS2-17236]: https://dhis2.atlassian.net/browse/DHIS2-17236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/developer/web-api/tracker.md
+++ b/src/developer/web-api/tracker.md
@@ -2924,6 +2924,12 @@ Tracker objects, the requested organisation unit must always be within the user'
 user cannot request objects outside its search scope unless it's using the organisation unit mode
 ALL and has enough privileges to use that mode.
 
+When requesting tracked entities without specifying a program, the response will include only those tracked entities that satisfy one of the following criteria:
+- The tracked entity is enrolled in at least one program the user has data access to, and the user has access to the owner organisation unit.
+- The tracked entity is not enrolled in at least one program the user has data access to, and the user has access to the tracked entity registering organisation unit.
+
+In both scenarios, the validation process will consider the program access level, as detailed at the beginning of this section, and the sharing settings, as explained [here](#webapi_tracker_metadata_sharing).
+
 #### Tracker Ownership Override: Break the Glass { #webapi_tracker_ownership_override }
 
 It is possible to temporarily override this ownership privilege for a program that is configured
@@ -3415,4 +3421,3 @@ All merging is based on data already persisted in the database, which means the 
 service is not validating that data again. This means if data was already invalid, it will not be
 reported during the merge. The only validation done in the service relates to relationships, as
 mentioned in the previous section.
-
--- a/src/developer/web-api/tracker.md
+++ b/src/developer/web-api/tracker.md
@@ -2924,11 +2924,9 @@ Tracker objects, the requested organisation unit must always be within the user'
 user cannot request objects outside its search scope unless it's using the organisation unit mode
 ALL and has enough privileges to use that mode.
 
-When requesting tracked entities without specifying a program, the response will include only those tracked entities that satisfy one of the following criteria:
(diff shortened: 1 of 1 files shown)


muilpp:
Since this affects TEs only, I wonder if it would make more sense to include it in the [Tracked Entities section](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker.html#tracked-entities-get-apitrackertrackedentities) instead. The drawback is that this would result in scattered ACL documentation.

src/developer/web-api/tracker.md
```diff
@@ -2924,6 +2924,12 @@ Tracker objects, the requested organisation unit must always be within the user'
 user cannot request objects outside its search scope unless it's using the organisation unit mode
 ALL and has enough privileges to use that mode.
 
+When requesting tracked entities without specifying a program, the response will include only those tracked entities that satisfy one of the following criteria:
+- The tracked entity is enrolled in at least one program the user has data access to, and the user has access to the owner organisation unit.
+- The tracked entity is not enrolled in at least one program the user has data access to, and the user has access to the tracked entity registering organisation unit.
```
teleivo: That sounds odd 🤔 

`The tracked entity is not enrolled in at least one program the user has data access to`

Can you phrase this differently? I don't I get it 😅

src/developer/web-api/tracker.md
```diff
@@ -2924,6 +2924,12 @@ Tracker objects, the requested organisation unit must always be within the user'
...
+
+In both scenarios, the validation process will consider the program access level, as detailed at the beginning of this section, and the sharing settings, as explained [here](#webapi_tracker_metadata_sharing).
```
teleivo: Can we merge this with 

> When requesting tracked entities without specifying a program, the response will include only those tracked entities that satisfy one of the following criteria:

to simplify things i.e. no reference back to the paragraph above.

How about

When requesting tracked entities without specifying a program, the response will include only tracked entities that satisfy [metadata sharing settings](#webapi_tracker_metadata_sharing) and one of the following criteria:

teleivo (commented):
I would keep the change in the ACL section. If you want to restructure the docs with regards to ACL we should rethink more than just this ACL addition.

src/developer/web-api/tracker.md
muilpp: it should be a bit more clear now

src/developer/web-api/tracker.md
```diff
@@ -2924,11 +2924,9 @@ Tracker objects, the requested organisation unit must always be within the user'
...
 ALL and has enough privileges to use that mode.
 
-When requesting tracked entities without specifying a program, the response will include only those tracked entities that satisfy one of the following criteria:
+When requesting tracked entities without specifying a program, the response will include only tracked entities that satisfy [metadata sharing settings](#webapi_tracker_metadata_sharing) and one of the following criteria:
 - The tracked entity is enrolled in at least one program the user has data access to, and the user has access to the owner organisation unit.
-- The tracked entity is not enrolled in at least one program the user has data access to, and the user has access to the tracked entity registering organisation unit.
-
-In both scenarios, the validation process will consider the program access level, as detailed at the beginning of this section, and the sharing settings, as explained [here](#webapi_tracker_metadata_sharing).
+- The tracked entity is not enrolled in any program the user has data access to, and the user has access to the tracked entity registering organisation unit.
```
```suggestion
- The tracked entity is not enrolled in any program the user has data access to, but the user has access to the tracked entity registering organisation unit.
```

teleivo (approved):

PR #1418, opened 2024-07-05
docs: Add effective search scope docs [DHIS2-17708]

muilpp:


src/developer/web-api/tracker.md
```diff
@@ -2906,6 +2906,23 @@ scope is applied to the registration organisation unit of the TrackedEntity. Whe
 retrieving TrackedEntities, including specific program data, the organisation unit scope is applied
 to the Owner organisation unit.
 
+In scenarios where the capture scope and search scope do not overlap, the system will grant the
```
teleivo: Do we even need to expose the user to this level of detail? And in what scenarios?

As a user I would find it easier to understand that in x, y, z the union of capture and search scope are used. And in a, b, c only the capture or search scope is used.

Can we make such statements? 😅

enricocolasante: I agree, I would not say anything about overlap and just saying that the user can search (or whatever we call the operations that can be done with the search scope) in all the orgunits present in Search Scope + Capture Scope.

Maybe we can just mention that the most common configuration is to have the Capture Scope as a subset of the Search Scope

muilpp: if we don't want that level of detail, then do you think one sentence like the one below would be enough?
`The search scope and capture scope units define the range within which a user can search for data.`

That's always true regardless of whether they overlap.

teleivo (approved):

PR #1419, opened 2024-07-10
docs: DataElement merge docs [DHIS2-14429]

david-mackessy:
Add docs for the `DataElement` merge feature.

### DIFF
--- a/src/developer/web-api/metadata.md
+++ b/src/developer/web-api/metadata.md
@@ -1597,7 +1597,7 @@ Table: Constraints and error codes
 | E1556      | dataMergeStrategy field must be specified. With value `DISCARD` or `LAST_UPDATED`                                                           |
 
 #### Database constraints
-If a database table has a unique key constraint, allowing only 1 unique data element as part of the constraint, then the merge will fail.  
+There are unique constraints in place that can prevent a successful merge. These constraints are set by DHIS2 in order to maintain a logical domain model.    
 Below are a list of the known database unique key constraints at the time of writing. For example, you
 can only have 1 data set element with the same dataset and data element.
 
--- a/src/developer/web-api/metadata.md
+++ b/src/developer/web-api/metadata.md
@@ -1488,17 +1488,21 @@ following payload to change the style:
 
 > **Caution**
 >
-> Merging DataElements should be carried out with the utmost care. Knowing
-> the potential side effects of a merge should be fully understood before performing
+> Merging DataElements should be carried out with the utmost care. Particular attention
+> should be given to the merging of data values that have data element references involved in the
+> merge. Knowing the potential side effects of a merge should be fully understood before performing
 > the merge. The merging of DataElements has far-reaching effects. The information below
 > will try to help show what's involved in a DataElement merge. A DataElement merge
-> touches all the major parts or the system (metadata, data, tracker and analytics).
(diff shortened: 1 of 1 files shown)


src/developer/web-api/metadata.md
```diff
@@ -1482,6 +1482,226 @@ following payload to change the style:
...
+| E1551      | Target data element must be specified                                                                                                       |
+| E1552      | Target data element cannot be a source indicator                                                                                            |
+| E1553      | Source/Target data element does not exist: `{uid}`                                                                                          |
+| E1554      | All source ValueTypes must match target ValueType: `ValueType`. Other ValueTypes found: `ValueType`                                         |
+| E1555      | All source DataElementDomains must match target DataElementDomain: `DataElementDomain`. Other DataElementDomains found: `DataElementDomain` |
+| E1556      | dataMergeStrategy field must be specified. With value `DISCARD` or `LAST_UPDATED`                                                           |
+
+#### Database constraints
+If a database table has a unique key constraint, allowing only 1 unique data element as part of the constraint, then the merge will fail.
```
jbee: This sounds unclear to me. Why would there be one or not be one? Who adds the constraint? Should one expect a constraint to be present? What is the situation where we get a collision?

I assume you mean this referring to the table that follows. I think we better state this as a given. Something like "There are unique constraints in place that can prevent a successful merge". When would this be a conflict compare to when does the merge work? I don't quite get the condition from this. What quality of the source or target is important here?

jbee (approved):
Very detailed, which is mostly good but we have to find a balance with docs as well. There is such a thing as too much or too detailed and you are approaching it :D

Some suggestions or questions on passages that are a bit unclear to me.

david-mackessy:
> Very detailed, which is mostly good but we have to find a balance with docs as well. There is such a thing as too much or too detailed and you are approaching it :D
> 
> Some suggestions or questions on passages that are a bit unclear to me.

@jbee What do you think could be stripped out? Possibly a sample json response? If I'm a user, I would want to know what's involved in this feature. I think most, if not all of this information is necessary. We want to help users understand the full implications of the feature, otherwise it would probably promote more incoming requests/bugs.

src/developer/web-api/metadata.md
david-mackessy: this information is all under the section `Database constraints`. So it is already logically grouped together. It outlines, in order:
- warning that DB constraints exist
- why they can occur
- when/where they can occur

I can update to say that these constraints are set by DHIS2 core team (this seems to be a very obvious detail though 😄 )
I can also update to say why a DB constraint might exist, which are basically domain constraints set by DHIS2.
And I can use your proposed alternative suggestion as well.

jbee:
> What do you think could be stripped out? Possibly a sample json response?

Yes, when it comes to responses we usually keep things unsaid also to give us some room to change details without needing a docs update. If we use response examples they should demonstrate something we wanted to highlight that we think isn't self-explanatory.

src/developer/web-api/metadata.md
jbee: The intro just sounded too loose where this might be up to users or operators/admins when I assume it is more correct to assume this is simply part of the software package.

david-mackessy: gotcha 👍

PR #1432, opened 2024-09-21
fix: update wildcard route documentation

kabaros:
the docs for wildcard mentioned that the URL should end with an asterisk `*`, but it's supposed to end with _two_ asterisks `**` otherwise the subpath doesn't work. Here is the [line](https://github.com/dhis2/dhis2-core/blob/59a289f6de2aaaf33fb3e199f07af97eefe61964/dhis-2/dhis-api/src/main/java/org/hisp/dhis/route/Route.java#L51C55-L51C57) of code defining this behaviour:

### DIFF
--- a/src/developer/web-api/route.md
+++ b/src/developer/web-api/route.md
@@ -102,14 +102,14 @@ Custom authorities allows users that don't have the metadata access to manage th
 
 ### Wildcard Routes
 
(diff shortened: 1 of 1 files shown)


amcgee (approved):
Thanks @kabaros !

PR #1441, opened 2024-10-22
doc: backport user login doc

netroms:
### Summary
Backports the new user login endpoint documentation.


tomzemp (approved):

jbee (approved):
As I understand our API maturity this would fall into _Beta_. Should we annotate the endpoints with `@Maturity.Beta` as well to make the same point visible in OpenAPI?

PR #1445, opened 2024-10-29
docs: Analytics table hooks in the Maintenance App [DHIS2-13473]

benguaraldi:
Adds basic documentation for analytics table hooks in the Maintenance section of the manual, pointing the reader to where to learn more.

### DIFF
--- a/src/user/configure-metadata.md
+++ b/src/user/configure-metadata.md
@@ -5097,7 +5097,7 @@ View".
 The Analytics Table Hooks functionality of DHIS2 stores SQL code that
 is run during different phases of the analytics table generation process.
 
-See also [the documentation on <code>/api/analyticsTableHooks</code>](../../../develop/using-the-api/dhis-core-version-master/analytics.html#webapi_analytics_table_hooks).
+See also [the documentation on <code>/api/analyticsTableHooks</code> in the Develop section of the manual](../../../develop/using-the-api/dhis-core-version-master/analytics.html#webapi_analytics_table_hooks).
 
 ### Creating a new analytics table hook
 
--- a/src/user/configure-metadata.md
+++ b/src/user/configure-metadata.md
@@ -34,6 +34,12 @@ need to collect and analyze data:
 
   - External map layers
 
+  - SQL views
+
+  - Locales
+
+  - Analytics table hooks
+
 > **Note**
 >
 > The functions you have access to depend on your user role's access
@@ -4984,7 +4990,7 @@ cultural region.
 
 5.  Click **Save**.
 
-## Manage SQL Views { #maintenance_sql_view } 
+## Manage SQL views { #maintenance_sql_view } 
 
 The SQL View functionality of DHIS2 will store the SQL view definition
 internally, and then materialize the view when requested.
@@ -5086,7 +5092,7 @@ View".
 > appears after view B in alphabetical order, analytics may fail, as the
 > view with dependencies will not be dropped in the correct order.
 
-## Manage Analytics Table Hooks { #maintenance_analytics_table_hooks } 
+## Manage analytics table hooks { #maintenance_analytics_table_hooks } 
 
 The Analytics Table Hooks functionality of DHIS2 stores SQL code that
 is run during different phases of the analytics table generation process.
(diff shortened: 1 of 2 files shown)


src/user/configure-metadata.md
```diff
@@ -5086,6 +5086,20 @@ View".
 > appears after view B in alphabetical order, analytics may fail, as the
 > view with dependencies will not be dropped in the correct order.
 
+## Manage Analytics Table Hooks { #maintenance_analytics_table_hooks } 
+
+The Analytics Table Hooks functionality of DHIS2 stores SQL code that
+is run during different phases of the analytics table generation process.
+
+See also [the documentation on <code>/api/analyticsTableHooks</code>](../../../develop/using-the-api/dhis-core-version-master/analytics.html#webapi_analytics_table_hooks).
```
jimgrace: Do we provide links between manuals in other places? If not, then I suggest we don't start here. They won't work in downloaded PDF versions of the docs. References to the Developer Manual that I see in other places are more like:
> You will find more information about the plug-ins in the Web API chapter in the *DHIS2 Developer Manual*

src/user/configure-metadata.md
```diff
@@ -5086,6 +5086,20 @@ View".
 > appears after view B in alphabetical order, analytics may fail, as the
 > view with dependencies will not be dropped in the correct order.
 
+## Manage Analytics Table Hooks { #maintenance_analytics_table_hooks }
```
jimgrace: I suggest "Manage analytics table hooks" for capitalization consistency with the other sections under "Configure metadata" (except for Manage SQL Views, the only outlier). I've noticed over the years that Americans tend to use uppercase more freely when naming things than do Europeans. (This is true also in the World Council of Churches!) DHIS2 documentation tends to follow more the European capitalization sensibilities.

jimgrace: Should add "Analytics table hooks" to the list of things you can configure under _Configure metadata_ / _About the Maintenance app_. This list is already missing:
- External map layers
- SQL views
- Locales

We should fix this oversight by adding these as well, not add to the oversight.

Note that "Program metadata: tracked entity, tracked entity attribute and relationship type" is in the list of things you can configure but is not a section under _Configure metadata_. This is quirky but OK because it has its own section _Configure programs in the Maintenance app_.

jimgrace (changes requested):
I suggest a note in the developer manual that you can configure Analytics table hooks in the Maintenance app as described in the user manual. This could go under _Analytics table hooks_ / _Creating hooks_, before describing how they can also be created using the WebAPI.

src/user/configure-metadata.md
benguaraldi: We could do that.

I have added links before.  Here are the two I could quickly find:
- https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/users.html#switching-between-user-accounts-connected-to-the-same-identity-provider-account
- https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/maintenance.html?h=means+latest+or+continuous+analytics+develop+master#webapi_generating_resource_analytics_tables

src/user/configure-metadata.md
benguaraldi: Yes, I used Manage SQL Views as my template, hence following that model.  I will fix this.

benguaraldi: External map layers was there, but the others were indeed missing.  I will add them.

benguaraldi:
Thanks, @jimgrace!  All changes except for removing the link have been implemented.

src/user/configure-metadata.md
jimgrace: Hmmm. Well you might reference the name of the document so that folks with a broken link (like the PDF) might still have an idea where to find it. It also might impart useful information to the user to know where the link goes without having to follow it. I often like to know where I'm going before I decide to go there! Perhaps something like:

> See also [the Developer Manual section on <code>/api/analyticsTableHooks</code>](../../../develop/using-the-api/dhis-core-version-master/analytics.html#webapi_analytics_table_hooks).

Also, you could consider an absolute URL which would work inside PDFs.

benguaraldi: Yes, that's a good idea to reference the name of the document.  I have done that in the most recent commit.

I did think the absolute url would be more brittle, but I'm not sure there's much argument for that.

Maybe let's submit this PR with the relative url (as it matches the others) and then I'll submit another PR that changes all of the relative urls to absolute urls and then submit that for @Philip-Larsen-Donnelly to consider?

jimgrace (approved):

PR #1447, opened 2024-10-30
docs: Change relative links to absolute links

benguaraldi:
In PR #1445, @jimgrace recommended that we consider absolute links instead of relative links in the documentation.  I have put together this PR to change the four relative links that I could find into absolute links.

One advantage of this approach is that it might work across more platforms.  For instance, if you're reading a PDF, the link would still work.  One disadvantage would be that it would switch the user between languages, from whatever language they are reading to English.

Maybe neither of these approaches makes sense, but it would be nice to have a way to link from one section of the documentation to another.

Requesting review from @Philip-Larsen-Donnelly, but I'm happy for you to pull in whoever else should be consulted.

### DIFF
--- a/src/developer/web-api/maintenance.md
+++ b/src/developer/web-api/maintenance.md
@@ -38,7 +38,7 @@ Table: Analytics tables optional query parameters
 > **Note**
 >
 > lastYears=0 means latest or continuous analytics, as defined in
-[Continuous analytics table](https://docs.dhis2.org/en/use/user-guides/dhis-core-version-master/maintaining-the-system/scheduling.html#scheduling_continuous_analytics_table).
+[Continuous analytics table](#scheduling_continuous_analytics_table).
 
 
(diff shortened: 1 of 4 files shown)


Philip-Larsen-Donnelly:
Neither relative or absolute references are reliable if the other sections are moved, and neither should be used. :sweat_smile:  


The correct (reliable) way to reference other parts of the documentation is to use unique anchors in the docs and refer to them, as described here: https://docs.dhis2.org/en/implement/support-and-documentation/dhis2-documentation-guide.html#name_of_section

benguaraldi:
Thanks, @Philip-Larsen-Donnelly!  I've changed the links to name_of_section links.  Do they look correct now?

jimgrace (approved):
Looks good, thanks!

benguaraldi:
Perhaps not surprisingly, but @Philip-Larsen-Donnelly's method works great!  (At least presuming the docs have been refreshed since yesterday.)

PR #1462, opened 2024-11-29
docs: update change logs params

muilpp:
Updates change logs docs according to new parameters and event properties feature.


src/developer/web-api/tracker.md
```diff
@@ -2139,6 +2139,8 @@ This endpoint retrieves change logs for the attributes of a specific tracked ent
 |---|---|---|
 |path `/{uid}`|`String`|Tracked entity `UID`.|
 |`program`|`String`|Program `UID` (optional).|
+|`order`|`String`|Name and sort direction pair in the format `propName:sortDirection`.<br><br>Example: `createdAt:desc`<br><br>`propName` is case-sensitive. Valid sortDirection values are `asc` and `desc`. `sortDirection` is case-insensitive and defaults to `desc`. Supported properties are `attribute`, `createdAt`, and `username`, only one at a time.|
+|`filter`|`String`|Colon-separated field name with operator and value in the format `fieldName:operator:value`.<br><br>Example: `attribute:eq:w75KJ2mc4zz`<br><br>Filtering is supported for `attribute` and `username` fields, one at a time. Only the `eq` (equals) operator is supported.|
```
teleivo: I think we should settle on one of property or field. Maybe we should go for `field` as `field` is already part of our APIs via the parameter `field`.

teleivo: if only `eq` is supported you can make the description less generic and state `fieldName:eq:value`

src/developer/web-api/tracker.md
```diff
@@ -2639,11 +2641,13 @@ although it might have multiple rows for each data element value.
 #### Event data value change logs { #webapi_event_data_value_change_logs }
 `GET /api/tracker/events/{uid}/changeLogs`
 
-This endpoint retrieves change logs for the data values of a specific event. It returns a list of all event data values that have changed over time for that particular event.
+This endpoint retrieves change logs for the data values of a specific event. It returns a list of all event data values and event properties (`occurredAt`, `scheduledAt`, and `geometry`) that have changed over time for the specified event.
 
 |Parameter|Type|Allowed values|
 |---|---|---|
 |path `/{uid}`|`String`|Event `UID`.|
+|`order`|`String`|Name and sort direction pair in the format `propName:sortDirection`.<br><br>Example: `createdAt:desc`<br><br>`propName` is case-sensitive. Valid sortDirection values are `asc` and `desc`. `sortDirection` is case-insensitive and defaults to `desc`. Supported properties are `createdAt`, `dataElement`, `property` and `username`, only one at a time.|
```
teleivo: Users can order by `property`? Not sure if it's clear what that is. Is that explained somewhere?

Will that also be supported in
`/api/tracker/trackedEntities/{uid}/changeLogs`

muilpp: That's explained here: https://github.com/dhis2/dhis2-docs/pull/1462/files#diff-4b5dabc4706dd8ae8c4c71cc43b6c8d08fa520c861a63cae3908ac407f201fffR2644

No, only event change logs support `property`

src/developer/web-api/tracker.md
```diff
@@ -1686,7 +1686,7 @@ The following endpoint supports standard parameters for pagination.
...
-|`order`|`String`|Comma-separated list of property name and sort direction pairs in format `propName:sortDirection`.<br><br>Example: `createdAt:desc`<br><br>Entities are ordered by newest (internal id desc) by default, meaning when no order parameter is provided.<br><br>**Note:** `propName` is case sensitive. Valid `sortDirections` are `asc` and `desc`. `sortDirection` is case-insensitive. `sortDirection` defaults to `asc` for properties or UIDs without explicit `sortDirection`.||
+|`order`|`String`|Comma-separated list of property name and sort direction pairs in format `propName:sortDirection`.<br><br>Example: `createdAt:desc`<br><br>Entities are ordered by newest (internal id desc) by default, meaning when no order parameter is provided.field `propName` is case sensitive. Valid `sortDirections` are `asc` and `desc`. `sortDirection` is case-insensitive. `sortDirection` defaults to `asc` for properties or UIDs without explicit `sortDirection`.||
```
```suggestion
|`order`|`String`|Comma-separated list of property name and sort direction pairs in format `propName:sortDirection`.<br><br>Example: `createdAt:desc`<br><br>Entities are ordered by newest (internal id desc) by default, meaning when no order parameter is provided. `propName` is case sensitive. Valid `sortDirections` are `asc` and `desc`. `sortDirection` is case-insensitive. `sortDirection` defaults to `asc` for properties or UIDs without explicit `sortDirection`.||
```

teleivo: Not sure if you wanted to use `field` or not 🤷🏻

src/developer/web-api/tracker.md
```diff
@@ -2139,8 +2139,8 @@ This endpoint retrieves change logs for the attributes of a specific tracked ent
 |---|---|---|
 |path `/{uid}`|`String`|Tracked entity `UID`.|
 |`program`|`String`|Program `UID` (optional).|
-|`order`|`String`|Name and sort direction pair in the format `propName:sortDirection`.<br><br>Example: `createdAt:desc`<br><br>`propName` is case-sensitive. Valid sortDirection values are `asc` and `desc`. `sortDirection` is case-insensitive and defaults to `desc`. Supported properties are `attribute`, `createdAt`, and `username`, only one at a time.|
-|`filter`|`String`|Colon-separated field name with operator and value in the format `fieldName:operator:value`.<br><br>Example: `attribute:eq:w75KJ2mc4zz`<br><br>Filtering is supported for `attribute` and `username` fields, one at a time. Only the `eq` (equals) operator is supported.|
+|`order`|`String`|Field and sort direction pair in the format `field:sortDirection`.<br><br>Example: `createdAt:desc`<br><br>`field` is case-sensitive. Valid sortDirection values are `asc` and `desc`. `sortDirection` is case-insensitive and defaults to `desc`. Supported fields are `attribute`, `createdAt`, and `username`, only one at a time.|
```
teleivo: Is the default intentionally different here than in the above endpoint

> `sortDirection` defaults to `asc`

🤔

src/developer/web-api/tracker.md
muilpp: Nope, that was a typo. 
Fixed now.

src/developer/web-api/tracker.md
muilpp: The endpoint above is not related to change logs, I just adapted it to the field naming for the sake of consistency. This one is about change logs, but what I was trying to say is a bit different.
Long story short, to make things easier, I'll explain the same thing as above, because they behave the same.

ameenhere (approved):

PR #1472, opened 2025-01-08
Update system-settings.md

karolinelien:
Removed "put analytics in maintenance mode" that somehow returned in docs, removed "threshold analytics data caching" which is also not in system settings app anymore, and added info on "enable embedded dashboards"


Philip-Larsen-Donnelly (approved):
LGTM

PR #1489, opened 2025-01-24
docs(DHIS2-18837): document new auth schemes

cjmamo:


src/developer/web-api/route.md
```diff
@@ -18,70 +18,105 @@ X-Forwarded-User: <username>
...
-GET /api/routes/postman-get/run
+GET /api/routes/postman/run
+POST /api/routes/postman/run
 ```
 
-### GET Route with authentication
+### Running a route with authentication
 
-Two types of authentication are supported: `http-basic` and `api-token`. In both cases they will simply add the required headers. In the case of `api-token` its meant to be used by our PAT (personal access token) so it might not be the right header for your purpose (the format is `ApiToken token`).
+A number of authentication modes are supported when running routes. These authentication modes add headers or query parameters to the request being routed from DHIS2. DHIS2 encrypts at rest sensitive headers or query parameters when creating a route with an authentication mode. This means that the unencrypted secrets cannot be read from the database or the Web API. What follows are the supported authentication modes:
```
david-mackessy: > DHIS2 encrypts at rest sensitive headers or query parameters when creating a route with an authentication mode. This means that the unencrypted secrets cannot be read from the database or the Web API.

Should it say `This means that the encrypted secrets cannot be read from the database or the Web API` ?
It reads as though **_unencrypted_** secrets have been saved to the DB otherwise.

cjmamo: > Should it say `This means that the encrypted secrets cannot be read from the database or the Web API` ?

Well, strictly speaking, the encrypted secrets can be read from the database. Maybe I should it rephrase it as:

`This means that the clear text secrets cannot be read from the database or the Web API.`

david-mackessy: Yeah I know what you mean 😀, It might be just how I'm reading it.
Your suggestion does make it clearer, for me anyway 👍

david-mackessy (approved):

netroms (approved):

jbee (approved):

src/developer/web-api/route.md
```diff
@@ -18,70 +18,105 @@ X-Forwarded-User: <username>
...
 
 ```json
 {
-  "name": "postman-get",
-  "code": "postman-get",
+  "name": "postman",
+  "code": "postman",
   "disabled": false,
-  "url": "https://postman-echo.com/get"
+  "url": "https://postman-echo.com/"
```
amcgee:
```suggestion
  "url": "https://postman-echo.com/get"
```

We should keep this pointing at postman-echo/get - the get endpoint echoes the request sent with a GET method, which is good for testing a route.  Using `postman-echo.com` without a path returns the HTML of the `postman-echo` website

src/developer/web-api/route.md
```diff
@@ -18,70 +18,105 @@ X-Forwarded-User: <username>
...
-### GET Route without authentication
+### Running a route
+
+The following is an example of a JSON request creating a route:
 
 ```json
 {
-  "name": "postman-get",
-  "code": "postman-get",
+  "name": "postman",
```
amcgee: (minor) in order to make it clear that the code (not the name) can be used to access the route, let's make this something different
```suggestion
  "name": "Postman Echo",
```

amcgee (approved):
Minor suggested changes, but otherwise looks good to me

PR #1490, opened 2025-01-28
docs: idScheme feature on /trackedEntities DHIS2-18540

teleivo:


src/developer/web-api/tracker.md
```diff
@@ -312,6 +312,50 @@ enrollment payload. Below is a sample payload:
...
+collections are always exported using `UID`s, except for:
+
+* `TrackedEntity.attributes[].attribute`
+* `Event.dataValues[].dataElement`
+
+For example, metadata references in `TrackedEntity.relationships` or `enrollments` will always use
+`UID`s for import/export.
+
+The [import](#webapi_tracker_import) expects metadata to use the chosen `idScheme`. There is no
+fallback - for example, trying `orgUnitIdScheme=CODE` first and then falling back to `UID` if not
```
ameenhere: I feel we need to make the example more clearer. 
The example says what "fallback" means. But better we be explicit in the "**no** fallback" in the example.

teleivo: Do you have a suggestion? I don't understand what you mean 😅

ameenhere: The most straightforward suggestion is to add a "is not possible." At the end of the example 😄  There may be other ways to make it explicit. 

"for example, trying `orgUnitIdScheme=CODE` first and then falling back to `UID` if not found is not possible"

Basically stating in the example there is no fallback.

Hope that was clearer? 😄

teleivo: I rephrased it. Maybe avoiding the fallback entirely makes it easier.

ameenhere (approved):

PR #1491, opened 2025-01-29
docs: added note about PRV newest event meaning

dayures:


src/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -1130,8 +1130,8 @@ objects:
...
+    | **Data element from the newest event for a program stage** | This source type works the same way as **Data element from the newest event in the current program**, except that it only evaluates values from one program stage.<br>     <br>This source type can be useful in program rules where the same data element is used in several program stages, and a rule needs to evaluate the newest data value from within one specific stage. <br>     <br>In order to know what event is the newest, the report date (event date) is used. Take into account that if you have many events with the same report date, the system choose one randomly.|
```
ameenhere: This may not be correct now. I think we have started using the time as well? @enricocolasante thoughts?

enricocolasante: Yes, we are not choosing randomly, we use the `createdAt` field for secondary order.

ameenhere: @dayures If you can modify both the statements to reflect what Enrico said, we can take this PR forward. 

In order to know what event is the newest, the report date (event date) is used. If you have many events with the same report date, the system choose the one with the latest `createdAt` property of the event.

src/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -1130,8 +1130,8 @@ objects:
...
+    | **Data element from the newest event in the current program** | This source type is used when a program rule variable needs to reflect the newest known value of a data element, regardless of what event the user currently has open.<br>     <br>This source type is populated slightly differently in **Tracker Capture** and **Event Capture** apps:<br>     <br>**Tracker Capture**: the program rule variable will be populated with the newest data value collected for the given data element within the enrollment.<br>     <br>**Event Capture**: the program rule variable will be populated with the current events data. <br>**NB** Future dates are "newer" than current or past dates. <br>     <br>In order to know what event is the newest, the report date (event date) is used. Take into account that if you have many events with the same report date, the system choose one randomly.|
```
ameenhere, line 1134: Same as above.

ameenhere: Same as above

ameenhere (changes requested):

src/user/configure-programs-in-the-maintenance-app.md
dayures: @ameenhere @enricocolasante thanks for the update. I am going to fix the change. In which version the use of createdAt as second filter was introduced, in order to include it in the doc?

enricocolasante: The fix was introduced in 2.41.2, so it will be present in 2.42 and it is present in master

src/user/configure-programs-in-the-maintenance-app.md
dayures: Change done

src/user/configure-programs-in-the-maintenance-app.md
dayures: @ameenhere change done.

enricocolasante (approved):

PR #1493, opened 2025-01-29
docs: Update documentation on event filters [DHIS2-15954]

muilpp:


muilpp:
done in https://github.com/dhis2/dhis2-docs/pull/1510

PR #1495, opened 2025-01-29
docs: Update documentation on event filters [DHIS2-15954][2.40]

muilpp:
Align event filter documentation to current behavior.
When filtering data elements, the operator and value are mandatory, while when filtering tracked entity attributes, they are optional.


src/developer/web-api/tracker.md
```diff
@@ -1591,8 +1591,8 @@ Returns a list of events based on the provided filters.
 |`program`|`String`|`uid`| Identifier of program|
 |`programStage`|`String`|`uid`| Identifier of program stage|
 |`programStatus`|`enum`| `ACTIVE`&#124;`COMPLETED`&#124;`CANCELLED`| Status of event in program |
-|`filter`|`String`|Comma separated values of data element filters|Narrows response to events matching given filters. A filter is a colon separated property or data element UID with optional operator and value pairs. Example: `filter=fazCI2ygYkq:eq:PASSIVE` with operator starts with `eq` followed by a value. Characters such as `:` (colon) or `,` (comma), as part of the filter value, need to be escaped by `/` (slash). Likewise, `/` needs to be escaped. Multiple operator/value pairs for the same property/data element like `filter=qrur9Dvnyt5:gt:70:lt:80` are allowed. Repeating the same data element UID is not allowed. User needs access to the data element to filter on it.|
```
teleivo, line 1594: why did you remove the `optional`, are they not optional?

muilpp: I removed it because this feature works from 2.41+ only.

teleivo: Ok, but it confused me due to your addition below which is

`A filter like `filter=H9IlTX2X6SL` returns all events where the given attribute has a value.`

thus saying operator/value pairs are optional.

Is it that the addition below is wrong and should be

`A filter like `filterAttributes=H9IlTX2X6SL` returns all events where the given attribute has a value.`

So the operator/value pair is optional for `filterAttributes`?

muilpp: exactly, it's optional for `filterAttributes` only

teleivo (approved):

PR #1498, opened 2025-02-06
DHIS2-18568: document response timeout setting and max body limitation

cjmamo:


src/developer/web-api/route.md
```diff
@@ -140,6 +130,22 @@ In the example shown below, we are configuring a route with `http-basic` authent
...
+
+The pre-configured response timeout for the origin when running a route is 10 seconds. A response taking longer than this will result in the DHIS2 server returning a gateway timeout error to the client. However, this timeout can be adjusted when creating or updating a route as demonstrated below:
+
+```json
+{
+  "name": "Postman Echo",
+  "code": "postman-post",
+  "disabled": false,
+  "url": "https://postman-echo.com/post",
+  "responseTimeout": 5
```
jbee: I guess it is a bit late now but the naming here isn't the best. For a timeout time given without any unit in the name I would always assume it is milliseconds. So any time that isn't milliseconds should have the unit in the name, e.g. `responseTimeoutSeconds`. Otherwise people have to find and read the documentation just to understand what number to put.

jbee (approved):

src/developer/web-api/route.md
cjmamo: It's not too late. I'll rename the property.

src/developer/web-api/route.md
```diff
@@ -140,6 +130,22 @@ In the example shown below, we are configuring a route with `http-basic` authent
...
+  "name": "Postman Echo",
+  "code": "postman-post",
+  "disabled": false,
+  "url": "https://postman-echo.com/post",
+  "responseTimeoutSeconds": 10
+}
+```
+
+The minimum permitted response timeout is 1 second while the maximum permitted timeout is 60 seconds. The `responseTimeoutSeconds` setting should be used with caution since concurrent, long-running routes could degrade DHIS2's overall performance.
+
```
amcgee, line 150: We should add a warning block quote here about the fixed timeouts in older versions and the need to assign a custom timeout for long-running routes after upgrading to 42.  We should also add that to the upgrade notes here : https://github.com/dhis2/dhis2-releases/tree/master/releases/2.42

src/developer/web-api/route.md
```diff
@@ -44,6 +32,8 @@ If you want DHIS2 to _POST_ the request to the route target, use the _POST_ HTTP
 POST /api/routes/postman/run
 ```
 
+For performance reasons, the maximum allowed size for a route response body is 256 kilobytes. A response body from the origin exceeding this limit will lead to the DHIS2 server returning a bad gateway error to the client.
```
amcgee: I think maybe we should also mention here the content-types that are (and are not) supported so far with the route API

amcgee (changes requested):
Requested a few minor additions but overall looks good

src/developer/web-api/route.md
cjmamo: There isn't a restriction on content-types as far as I can tell.

src/developer/web-api/route.md
```diff
@@ -44,6 +32,8 @@ If you want DHIS2 to _POST_ the request to the route target, use the _POST_ HTTP
 POST /api/routes/postman/run
 ```
 
+For performance reasons, the maximum transfer time for a route response is 5 minutes. A response body from the upstream server exceeding this limit will lead to the DHIS2 server returning a bad gateway error to the client.
```
amcgee: Let's move this down to the timeout section, it's a bit out of context here and confusing since we also talk about a response timeout below.

amcgee (approved):
LGTM - let's just move the global max time line down to the response timeout section.

There are some general clarity improvements we should make here but not related to the timeout so will create a separate PR

PR #1499, opened 2025-02-10
Docs: Analytics - updated app list

gassims:
Updated app list so that instead of listing deprecated apps. It's now "... in the bundled analytics apps such as the **Data Visualizer app**." (cc: @janhenrikoverland)

Thanks!


gassims (commented):
Looks great, thanks!

Philip-Larsen-Donnelly (approved):

PR #1523, opened 2025-03-18
docs: Remove system id from dhis conf

david-mackessy:
It is not possible to set the system ID through the use of `dhis.conf`, so removing it from the docs.

The system ID is configured on startup by retrieving the `systemid` column value from the `configuration` table.
If it is null, a random `UUID` is then assigned (see [ConfigurationPopulator](https://github.com/dhis2/dhis2-core/blob/bcf087c2d7f26a92056b67c3ca66f6c579792a40/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/startup/ConfigurationPopulator.java))

If a different value is required, it could be manually updated directly through the DB or through the configuration API (when a restart is required before the updated value can be seen).

### DIFF
--- a/src/sysadmin/installation.md
+++ b/src/sysadmin/installation.md
@@ -1792,9 +1792,6 @@ server.https = off
 # System [Optional]
 # ----------------------------------------------------------------------
 
-# System identifier
-system.id = hmis1.country.org
-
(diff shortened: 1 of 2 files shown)


larshelge (approved):

larshelge:
Good catch.

PR #1528, opened 2025-03-21
feat: adding a note explaining userroles available

dayures:
Based on https://community.dhis2.org/t/user-cannot-access-roles-when-creating-another-user/2984/2 from @larshelge

### DIFF
--- a/src/user/manage-users-user-roles-and-user-groups.md
+++ b/src/user/manage-users-user-roles-and-user-groups.md
@@ -225,6 +225,12 @@ names or organisation unit level names.
 
 6. In the **Available roles** section, double-click the user roles you want to assign to the user.
 
+> **Note**
+>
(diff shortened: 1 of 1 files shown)


jason-p-pickering (changes requested):
Thanks @dayures . I think we should go a bit further here. The important concept is that users 
1) Should not be able to give away their own roles due to the security issues you mention
2) Only grant other users roles which themselves are a subset of their own role/roles. 

While I think we can point out how to enable users giving away their own roles, I think we should provide a much clearer warning about what you generally do not want to do this.

dayures:
@jason-p-pickering what do you think about adding this line at the beginning? "By default, to prevent security issues, users are not be able to give away their own roles. Users should only be able to grant other users roles that are a subset of their own."

PR #1529, opened 2025-03-21
docs: Add example to relationship model [DHIS2-19169]

enricocolasante:
Adding an example to better explain the relationship model and how we name things


src/user/relationship-model.md
```diff
@@ -71,3 +71,15 @@ One important thing to note about bidirectional relationships, are that the
 "from" and "to" sides are still significant in the database, meaning each entity
 must match the constraint for that side. However, from a user perspective, which
 side each entity is stored as is insignificant.
+
+## Example { #relationship_model_example }
+
+<img src="resources/images/relationship_model/relationships.png" width="50%"/>
```
teleivo: Did you choose the `img` tag to specify the `width`? This seems to be the first usage. Other images are added like

https://github.com/dhis2/dhis2-docs/commit/156874b891f2b996061aca7bcf9b9dbd30003a47

https://docs.dhis2.org/en/use/user-guides/dhis-core-version-master/working-with-your-account/messaging.html?h=define+who+you+want+to+receive+the+message+master#create-a-message

src/user/relationship-model.md
```diff
@@ -71,3 +71,15 @@ One important thing to note about bidirectional relationships, are that the
...
+
+* Bidirectional relationship *relA* connects *teA* to *teB* and connects *teB* to *teA*.
+* Unidirectional relationship *relB* connects *teA* to *teC* but DO NOT connect *teC* to *teA*.
```
teleivo:
```suggestion
* Unidirectional relationship *relB* connects *teA* to *teC* but DOES NOT connect *teC* to *teA*.
```

src/user/relationship-model.md
```diff
@@ -71,3 +71,15 @@ One important thing to note about bidirectional relationships, are that the
 "from" and "to" sides are still significant in the database, meaning each entity
 must match the constraint for that side. However, from a user perspective, which
 side each entity is stored as is insignificant.
+
+## Example { #relationship_model_example }
```
teleivo, line 75: Not sure how exactly but it would be great if that image would integrate with all of the text above. I think that would make the text easier to understand. Also an image where the relationships are nodes with from/to edges would help.

teleivo (commented):
Did you build the docs and check how it looks? Do we need the png and svg in the codebase? One image + the source code to generate it would be enough.

Philip-Larsen-Donnelly (approved):
This is good enough to include now, even if it could be enhanced in the future

PR #1567, opened 2025-05-05
chore: add node.primary_leader

dayures:
Key introduced https://github.com/dhis2/dhis2-core/commit/f5b4e5c95740fbec08085b4b22ddd2f916878003

### DIFF
--- a/src/sysadmin/reference/dhis.conf.md
+++ b/src/sysadmin/reference/dhis.conf.md
@@ -169,6 +169,9 @@ ldap.search.filter = (cn={0})
 # Node identifier, optional, useful in clusters
 node.id = 'node-1'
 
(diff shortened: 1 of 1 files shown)


ameenhere (approved):

ameenhere:
@dayures Thanks for the PR. Would be good to have this for v42 branch as well as the key was introduced in that version.

dayures:
@ameenhere sorry, I thought that commited that in master implies that it is commited in v42. Should I do for every change in master?

ameenhere:
> @ameenhere sorry, I thought that commited that in master implies that it is commited in v42. Should I do for every change in master?

@dayures It was like that until today I believe. Since v42 is being released soon, we are preparing for the change by branching out `2.42` from `master`. 

I think I merged this PR to master after the 2.42 branch was branched out. 😄 

But you can check if your changes are also available in 2.42 branch, if yes, then no need to do anything

PR #1571, opened 2025-05-06
chore: add analytics database keys (from source code)

dayures:


dayures:
@jason-p-pickering could you review this?

jason-p-pickering:
will defer to @luciano-fiandesio

bobjolliffe (changes requested):
Great to see these keys coming in to the documentation.  I would recommend commenting out the actual settings in this case.  Sometimes people will take the reference conf file and modify it (I sometimes do that).  In that case they might miss to comment these out.  By default there should be no analytics.database settings.  Unless you do indeed have another separate postgresql database for this

dayures:
Thanks for the feedback @bobjolliffe . I am happy to contribute in the most useful way. BTW, there is a big note before the file content saying "You should not attempt to use this configuration file directly, rather use it as a reference for the available configuration options. Many of the properties are optional."

How should be added the "analytics.database" parameters? Adding them commented out ?

PR #1574, opened 2025-05-08
docs: Program cat. mappings WebAPI [DHIS2-18745]

jimgrace:
This adds a new metadata section "Programs" to the Developers Manual (just before "Program Rules"). After a short introduction it has one subsection "Program Category Mappings" which describes how to add program category mappings using the Web API (copying some text from [DHIS2-18745](https://dhis2.atlassian.net/browse/DHIS2-18745) "The making of the demo").

This may be useful for sites that want to programmatically generate category mappings instead of using the frontend. For example, some sites are known to have thousands of attribute category options in a single category.

[DHIS2-18745]: https://dhis2.atlassian.net/browse/DHIS2-18745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/developer/web-api/metadata.md
```diff
@@ -2780,6 +2780,143 @@ resource for a predictor:
...
+In the Web API this looks like:
+
+```json
+"categoryMappingIds": [],
+```
+
+You can replace this with the category mappings that you want the program indicator to use.
+For example, if the program indicator has selected a category combination that combines
+`Gender` and `Outcome`, this field could be edited to contain the mapping ids for these
+categories such as defined in the above `categoryMppings` example:
```
DavidCKen:
```suggestion
categories such as defined in the above `categoryMappings` example:
```

DavidCKen (approved):
Looks good, just one little typo

PR #1577, opened 2025-05-13
docs: route manager app

kabaros:
add a link from the Route API page to the Route Manager app user guide


kabaros:
I added the Route Manager app docs to the Route API page - I don't think it belongs with "other apps" section. That section - other than being outdated - contains apps that are more user-facing. I think the entry point for this specific topic is the Route API - people come for the API, and here we tell them there is a helper UI for managing that API. I don't think there is a need for another place, not even a link ... happy to hear your thoughts @KaiVandivier @amcgee 

image

kabaros:
the app documentation is now published, so this PR is ready and the link here is ready to merge

KaiVandivier (approved):

PR #1603, opened 2025-06-27
doc: Add note on PR expression Date format should always be 'YYYY-MM-DD'

ameenhere:
Arising from this https://dhis2.atlassian.net/browse/DHIS2-15951 , all absolute dates in program rule expressions should have the format 'YYYY-MM-DD' irrespective of the system date format setting. Adding a note in the documentation for this.


linadhis2 (approved):
LGTM

PR #1608, opened 2025-07-04
chore: add hibernate keys (from source code)

dayures:


larshelge:
We don't want this to be changed.

dayures:
@larshelge why? I mean, if that are paremeters on the dhis.conf file, why shouldn't be documented?

PR #1611, opened 2025-07-08
DHIS2-13818: Deprecate C3P0 database connection pool

cjmamo:


jbee (commented):
Doesn't hikari share a lot of the settings with c3po? Are all these now deprecated ones really c3po only?

cjmamo:
>Doesn't hikari share a lot of the settings with c3po? Are all these now deprecated ones really c3po only?

It shares a few settings with C3P0. Only the C3P0-specific ones have been deprecated but you can see how the docs are confusing since it doesn't say which settings are C3P0-specific.

david-mackessy (approved):

PR #1612, opened 2025-07-09
feat: add doc for connection pool settings

netroms:


cjmamo:
While AI-generated content may help reduce the workload, we should be sure that such content is double-checked by the PR author. There were quite a few mistakes especially when it came to version compatibility that forced me to double-check everything. In one case, the parameter name was incorrect.

cjmamo (approved):

david-mackessy (approved):

PR #1628, opened 2025-07-23
docs: Document min characters & blocked operators [DHIS2-12243]

muilpp:
Documenting the use of minimum characters to search and blocked operators on a TEA.
I will add a description of what `minimumCharactersToSearch`, `preferredOperator` and `blockedOperators` do in OpenAPI.

I'm also removing non tracker operators from the docs.

### DIFF
--- a/src/developer/web-api/tracker.md
+++ b/src/developer/web-api/tracker.md
@@ -1887,7 +1887,9 @@ Supported binary operators:
 
 *These operators are currently supported but may be removed in the future. We recommend using the operator mentioned in the description, as it provides the same functionality.
 
-Matches are case-insensitive. For example, `filter=w75KJ2mc4zz:eq:Scott` would return values of the given attribute that match any variation of "Scott" regardless of case, such as SCOTT, scott, Scott...
+Matches are case-insensitive, for example `eq` and `ieq` (`i` for `insensitive`) behave in the same way. To ensure future compatibility, always use the non-i form (eq, like, etc.).
+
+For instance, `filter=w75KJ2mc4zz:eq:Scott` would return values of the given attribute that match any variation of "Scott" regardless of case, such as SCOTT, scott, Scott...
 
 Supported unary operators:
 
--- a/src/developer/web-api/tracker.md
+++ b/src/developer/web-api/tracker.md
@@ -1765,7 +1765,7 @@ The endpoint returns a list of tracked entities that match the request parameter
 
 |Request parameter|Type|Allowed values|Description|
 |---|---|---|---|
-|filter|String|Comma separated values of attribute filters.|Narrows response to tracked entities matching given filters. A filter is a colon separated property or attribute UID with optional operator and value pairs. Example: `filter=H9IlTX2X6SL:sw:A` with operator starts with `sw` followed by a value. A filter like `filter=H9IlTX2X6SL:!null` returns all events where the given attribute has a value. Special characters like `+` need to be percent-encoded, so `%2B` instead of `+`. Characters such as `:` or `,`, as part of the filter value, need to be escaped by `/`. Likewise, `/` needs to be escaped. Multiple operators for the same attribute like `filter=AuPLng5hLbE:gt:438901703:lt:448901704` are allowed. |
+|filter|String|Comma separated values of attribute filters.|Narrows response to tracked entities matching given filters. More on filters [here](#tracked_entity_attribute_filtering) |
 |orgUnits|String|Comma-separated list of organisation unit `UID`s.|Only return tracked entities belonging to provided organisation units|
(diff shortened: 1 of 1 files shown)


src/developer/web-api/tracker.md
```diff
@@ -1869,30 +1869,47 @@ Supported binary operators:
...
-| ieq | equal to, ignoring case |
 | ge | greater than or equal to (uses integer/number semantics for integer/decimal value types) |
 | gt | greater than, uses integer/number semantics for integer/decimal value types |
 | le | less than or equal to, uses integer/number semantics for integer/decimal value types |
-| lt | less than (uses integer/number semantics for integer/decimal value types |
-| ne | not equal to (uses integer/number semantics for integer/decimal value types |
-| neq | not equal to (uses integer/number semantics for integer/decimal value types |
-| nieq | not equal to |
+| lt | less than (uses integer/number semantics for integer/decimal value types) |
+| ne | not equal to (uses integer/number semantics for integer/decimal value types) |
```
enricocolasante, line 1877: Something like `filter=w75KJ2mc4zz:neq:Scott` is going to return a 400?
And what about `nlike`?

And what something like `ieq`?

I just don't understand if they are not supported or they are automatically translated to something else.
If they are not supported, then it is correct that they are not in the list, if they are supported but they are behaving in a different way, then they should be in the list and we should explain what is the behavior

ameenhere: Good point @enricocolasante 
I guess as it stands, it is supported but both of them behave in the case insensitive way. 
Perhaps we can document exactly as is working as of now. Removing the completely unsupported ones and merging the case insensitive and sensitive variant as one row? @muilpp  What do you think?

muilpp: All of them work as of now, the I-variant operators, `neq`, and `nlike`.

The reason I removed them from the list is that these operators are, in theory, specific to analytics. I say in theory because that was their original purpose, but since we share the same class, they can also be used in Tracker.

Since it's currently not possible to split the operators into Tracker-specific and Analytics-specific groups, I thought removing them from the Tracker docs would be a good first step to stop promoting their usage.

That said, if you think it makes more sense to document the behavior, given that the API does support it, should I find a way to include them while still discouraging their use?

enricocolasante: I think so, if it will not through a validation error then it should be on the list.
Then in the description we could say that the behavior is the same of the operator without the `i` and to use that operator instead.

src/developer/web-api/tracker.md
```diff
@@ -1869,30 +1869,55 @@ Supported binary operators:
...
+| neq | not equal to (uses integer/number semantics for integer/decimal value types), use `ne` instead* |
+| nieq | not equal to, ignoring case (use `ne` instead)* |
 | in | one of multiple values separated by semicolon ";", uses integer/number semantics for integer/decimal value types |
 | like | like text match |
-| ilike | like text match, ignoring case |
-| nilike | not like |
-| nlike | not like |
+| ilike | like text match, ignoring case (use `like` instead)*|
+| nilike | not like, ignoring case (use `like` instead)* |
```
ameenhere: Do we support `nlike` ?  If we do it is not same as `like` , it has to be the opposite

muilpp: we do support `nlike`, but it's part of the analytics operators, that's why I'd recommend using `like` instead.
would it be better if I write: _(use `like` instead if possible)_ ?

ameenhere: What I meant was, nlike and like are supposed to have the opposite behaviour right? Saying "use like instead" for nlike makes it sound like they both are interchangeable operators having the same behaviour. 

If we support nlike, lets just not say anything special for it. 
For nilike, we could say "use nlike instead". Wdyt?

src/developer/web-api/tracker.md
```diff
@@ -1869,30 +1869,55 @@ Supported binary operators:
...
+| nlike | not like, ignoring case (use `like` instead)* |
 | sw | starts with |
 | ew | ends with |
 
-Matches are case-insensitive, for example `eq` and `ieq` (`i` for `insensitive`) behave in the same way.
+*These operators are currently supported but may be removed in the future. We recommend using the operator mentioned in the description, as it provides the same functionality.
+
+Matches are case-insensitive. For example, `filter=w75KJ2mc4zz:eq:Scott` would return values of the given attribute that match any variation of "Scott" regardless of case, such as SCOTT, scott, Scott...
```
ameenhere:
```suggestion
Matches are case-insensitive, for example `eq` and `ieq` (`i` for `insensitive`) behave in the same way. To ensure future compatibility, always use the non-i form (eq, like, etc.).

For instance, `filter=w75KJ2mc4zz:eq:Scott` would return values of the given attribute that match any variation of "Scott" regardless of case, such as SCOTT, scott, Scott...
```

ameenhere (approved):

PR #1630, opened 2025-08-12
doc: update 2.42 server requirements (index.md, linux-manual-install.md)

tkipkurgat:
chore: fix sysadmin_section_index.yml to reference master


Philip-Larsen-Donnelly (approved):
Looks good. The only thing, if I was being pedantic, is that we put JDK as a requirement for runtime when it is actually JRE.

PR #1674, opened 2025-11-19
Add images for PI disaggregation section

DavidCKen:


Philip-Larsen-Donnelly (commented):
Some of these images look like they are at a higher resolution than we recommend for the docs site (1440x900): 
https://docs.dhis2.org/en/implement/support-and-documentation/dhis2-documentation-guide.html#taking-screenshots

DavidCKen:
re-adding files with correct res

PR #1688, opened 2025-12-17
docs: Add operators that can be blocked [DHIS2-20464]

muilpp:
Adding to docs the operators that can be blocked.


jason-p-pickering (dismissed):
In the section above you mention the use of the operator `sw` but here you use `SW`.  Are these separate operators and is this clear to the user?

muilpp:
> In the section above you mention the use of the operator `sw` but here you use `SW`. Are these separate operators and is this clear to the user?

Updated this section to use lowercase.

enricocolasante (approved):

ameenhere (approved):

PR #1711, opened 2026-01-25
More Content on Docker and Kubernetes install

tkipkurgat:
More content on docker and Kubernetes install Section.

### DIFF
--- a/src/sysadmin/index.md
+++ b/src/sysadmin/index.md
@@ -14,95 +14,203 @@ This chapter provides a guide for setting up the above technology stack. It shou
 
 ---
 
-DHIS2 is a database intensive application and requires that your server has an appropriate amount of RAM, a good number of CPU cores and a fast disk, preferably an solid state disk (ssd). These recommendations should be considered as rules-of-thumb and not exact measures. DHIS2 scales linearly on the amount of RAM and number of CPU cores so the more you can afford, the better the application will perform.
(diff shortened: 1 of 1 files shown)


src/sysadmin/index.md
```diff
@@ -14,95 +14,204 @@ This chapter provides a guide for setting up the above technology stack. It shou
 
 ---
 
-DHIS2 is a database intensive application and requires that your server has an appropriate amount of RAM, a good number of CPU cores and a fast disk, preferably an solid state disk (ssd). These recommendations should be considered as rules-of-thumb and not exact measures. DHIS2 scales linearly on the amount of RAM and number of CPU cores so the more you can afford, the better the application will perform.
+DHIS2 is a database-intensive application that scales linearly with RAM and CPU resources. These recommendations should be considered starting points - more resources will improve performance.
```
bobjolliffe, line 17: Agree this reads better.  Ideally we should get rid of this section entirely or restructure it substantially.  The narrative of a small, medium or large instance on a VM has too many assumptions baked in and increasingly doesn't reflect what people do.  For example provisioning postgresql separately.  Cloud vs datacentre dynamics etc.  Also there is not much of a clue what small, medium or large even implies.  If it is about guiding provisioning and budgeting there should also be mention of backup resources, staging, training etc.  So for the moment we can accept the improvement but lets rethink a bit here.

bobjolliffe (approved):

PR #1745, opened 2026-04-30
fix: Typo and grammar in configure programs

ameenhere:


src/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -2034,7 +2034,7 @@ There are two limits that can be set for a TET search
 To be able to search in other organisation units than the users data
 capture organisation units, the user must be assigned with search
 organisation units. Giving a user a search organisation unit will also
-give it access to search in all children of that organisation unit.
+give them access to search in all children of that organisation unit.
```
enricocolasante, line 2037: Wasn't correct before?

src/user/configure-programs-in-the-maintenance-app.md
```diff
@@ -1751,13 +1751,13 @@ For more information about configuration and the meaning of 'From constraint' an
...
 
-A tracked entity is a types of entities which can be tracked through the
+A tracked entity is a type of entity which can be tracked through the
 system. It can be anything from persons to commodities, for example a
 medicine or a person.
 
 A program must have one tracked entity. To enroll a tracked entity
-instance into a program, the tracked entity type and tracked
-entity type of a program must be the same.
+instance into a program, the tracked entity type of the instance and
```
enricocolasante:
```suggestion
instance into a program, the tracked entity type of the entity and
```

I would rather not use the word `instance`. This word, hopefully, does not have this meaning for users anymore.

src/user/configure-programs-in-the-maintenance-app.md
ameenhere: The subject is a "user" so "them" is better than "it" right?

enricocolasante (approved):

PR #1775, opened 2026-07-22
docs: update data administration guide for DHIS2-20439

DavidCKen:
Updates the Data Administration user guide to match current app behaviour, closing gaps identified in DHIS2-20439:

- Data integrity checks: documents the Standard/Slow check tabs and that slow checks must be run individually.
- Analytics tables management: adds the two new skip options (tracked entity data, outlier data) and reorders the existing options to match the app's actual order.
- Lock exceptions: adds sections on deleting a single lock exception and batch deletion (by data set + period, across all org units).
- Data Statistics: refreshed screenshot reflecting current object types (indicator groups/types, indicators, maps, user groups, users).

Targets master only for now — the underlying app features already shipped in earlier releases (2.37, 2.41), so the same doc gap likely exists on published version branches too; backporting docs needs to be followed up on a seperate PR with different screenshots of the data statistics page as Active Users only applies from 43 https://dhis2.atlassian.net/browse/DHIS2-20482

Closes DHIS2-20439.


Philip-Larsen-Donnelly (approved):
LGTM

PR #1777, opened 2026-08-06
docs: fix metadata attribute value filter example for 2.42+

netroms:
The metadata API filtering docs still show the pre-2.42 example:

```
/api/dataElements.json?filter=attributeValues.attribute.id:eq:n2xYlNbsfko&filter=attributeValues.value:eq:AFP
```

Since 2.42, metadata attribute values are stored as a JSONB map keyed by attribute ID (DHIS2-17482), so the `attributeValues.value` path no longer resolves and returns `400 E1003 Unknown path property`. The supported syntax uses the attribute's ID as the property name, which the same page already describes in the filter intro:

```
/api/dataElements.json?filter=n2xYlNbsfko:eq:AFP
```

This updates the example accordingly and adds an example for matching any value with `:!null`.

Reported on Slack by a community member hitting the 400 on v42+. Needs cherry-pick to `2.43` and `2.42`.

AI Assisted

### DIFF
--- a/src/developer/web-api/metadata.md
+++ b/src/developer/web-api/metadata.md
@@ -325,11 +325,15 @@ query using the id property of the associated data element groups:
 
     /api/dataElements.json?filter=dataElementGroups.id:eq:qfxEYY9xAl6
 
-To get data elements with a particular attribute value for a metadata 
-attribute, a filter for the attribute ID and the attribute value can be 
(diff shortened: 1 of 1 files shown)


david-mackessy (approved):

david-mackessy:
it's not very intuitive - wonder was this intentional or does it just work this way by accident?

PR #1778, opened 2026-08-06
docs: fix metadata attribute value filter example for 2.42+ [2.43]

netroms:
Cherry-pick of #1777 to `2.43`.

Updates the stale pre-2.42 `attributeValues.value` filter example (returns `400 E1003` since DHIS2-17482) to the supported attribute-ID-as-property-name syntax.

AI Assisted


david-mackessy (approved):

cjmamo:
The `.json` suffix is redundant yeah?

PR #1781, opened 2026-08-12
Adding Acronym page

DavidCKen:
Adding an acronym page to accompany the glossary to help disambiguate any acronyms that are specific for DHIS2 and help train AI agents on the specific terms that are used around HISP which is ironically not a acronym but like to use them a lot. Future use of this list might extend to a slack bot that could be called to answer acronym questions directly if needed. While loading the page here, the place in the navigation should be somewhere that is not version controlled and the current glossary should probably be moved out a version controlled menu as well


Philip-Larsen-Donnelly (approved):
Nice!

PR #1785, opened 2026-08-27
docs: document period type label and displayLabel [DHIS2-20845]

karolinelien:
Documents the `label` / `displayLabel` support on `PeriodType` added in 2.43, following up on https://github.com/dhis2/dhis2-releases/pull/261.

Adds a `Period types` subsection to the Web API introduction chapter (`src/developer/web-api/overview.md`), inside the existing *Date and period format* section and just after the period-format table, since that is where period types are already described. Covers `GET /api/periodTypes` with a property table marking `label` as the only writable field, plus a `Custom period type labels` subsection with the `PUT` request, the success response, the verify response, and clearing a label with an empty string.

Three details were checked against `dhis2-core` rather than taken from the release note, and differ from it:

- The release note's example body includes all fields (`isoDuration`, `frequencyOrder`, and so on). `PeriodTypeController.putPeriodType` only reads `getName()` and `getLabel()`, so everything else is silently ignored. The docs show `name` + `label` only.
- An unknown `name` returns `400 Bad Request` with the message `"<name> does not exist."` — `DefaultPeriodService.updatePeriodTypeLabel` throws `IllegalArgumentException`, which `CrudControllerAdvice` maps to `badRequest`. The status was not stated anywhere.
- `displayLabel` is `null`, not `displayName`, when no label is set — `PeriodType.getDisplayLabel()` returns `label` directly (with a `TODO` for the real implementation). The section therefore tells clients to render `displayLabel` and fall back to `displayName`, so apps adopting `displayLabel` now do not show blanks.

Also documents the `ALL` authority requirement (`@RequiresAuthority(anyOf = ALL)`).

Targets `master` only. If the 2.43 docs should carry it as well, that is a cherry-pick onto the `2.43` branch.

AI Assisted

### DIFF
--- a/src/developer/web-api/overview.md
+++ b/src/developer/web-api/overview.md
@@ -509,6 +509,104 @@ Table: Period format
 | Financial Year July | yyyyJuly | 2004July | July 2004-June 2005 |
 | Financial Year Oct | yyyyOct | 2004Oct | Oct 2004-Sep 2005 |
 
+### Period types { #webapi_period_types }
+
+The period types available in DHIS2 can be listed from the following
+resource:
+
+    GET /api/periodTypes
+
+The `fields` query parameter is supported and defaults to all fields.
+
+Table: Period type properties
+
+| Property | Description | Writable |
+|---|---|---|
+| name | Internal, hard-coded name of the period type, e.g. `FinancialFeb`. Identifies the period type. | No |
+| displayName | Translated, human-friendly name of the period type. | No |
+| isoDuration | Duration of the period type in ISO 8601 notation, e.g. `P1Y`. | No |
+| isoFormat | Format used for period identifiers of this type, as listed in the table above, e.g. `yyyyFeb`. | No |
+| frequencyOrder | Approximate number of days covered by the period type, used for ordering period types by frequency. | No |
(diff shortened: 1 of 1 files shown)


src/developer/web-api/overview.md
```diff
@@ -509,6 +509,104 @@ Table: Period format
...
+  "isoFormat": "yyyyFeb",
+  "frequencyOrder": 365,
+  "label": "Academic Year (February)",
+  "displayLabel": "Academic Year (February)"
+}
+```
+
+To remove a custom label, set `label` to an empty string:
+
+```json
```
maikelarabori, line 599: You can also do:

```json
{
  "name": "FinancialFeb",
  "label": null
}
```

maikelarabori (approved):
Looks good. Just one minor comment.

Thx!

src/developer/web-api/overview.md
karolinelien: Replaced with null but Claude also added a little note on empty vs null so have a look on whether that makes sense too :)

maikelarabori (approved):

src/developer/web-api/overview.md
maikelarabori: It looks good to me

PR #1786, opened 2026-08-28
fix: import/export guide clean up [DHIS2-20441]

tomzemp:
Some clean up based on comments in https://dhis2.atlassian.net/browse/DHIS2-20441

I am not updating the screen shots in this PR as we should arguably update all screenshots, and I think that would be better saved until our work with import/export app updates is completed.


Sharmyn28 (approved):
Thanks Tom! 😄

PR #1788, opened 2026-09-01
docs: retire Event Capture and Tracker Capture user documentation

karolinelien:
## Summary

The **Event Capture** and **Tracker Capture** apps were retired in DHIS2 2.42 ([community announcement](https://community.dhis2.org/t/retirement-of-the-old-data-entry-app-and-the-tracker-capture-app-in-v42/64852)), and all of their functionality now lives in the **Capture** app. The user guide still documented both as current apps.

> [!IMPORTANT]
> Needs https://github.com/dhis2/dhis2-docs-builder/pull/287 to merge **first, or alongside**. That PR drops the `master`-pinned Event Capture nav entries; without it the master user guide build fetches a file this PR deletes.

## What changed

**Event Capture** — chapter and its 2 images removed outright. Nothing linked to any of its 15 anchors.

**Tracker Capture** — chapter replaced with an 11-line placeholder (from 832) and 16 images removed. The placeholder:
- keeps the `{ #tracker_capture_app }` anchor, so inbound links from other versions and repos still land somewhere useful;
- points at the Capture app via `#capture_app`, and links the [Tracker Capture to Capture](https://docs.dhis2.org/en/implement/configuration/transition-kits/tracker-capture-to-capture.html) transition kit;
- links the 2.41 docs for anyone still on an older version.

That last link is a **full URL rather than an anchor** on purpose. The builder ignores anchors belonging to other versions when the linking page is version-specific, so `#tracker_capture_app` from master would resolve straight back to this placeholder instead of reaching 2.41.

**Retired authority removed** — `src/user/scheduling.md` told readers to use the authority `Ignore validation of required fields in Tracker and Event Capture` (`F_IGNORE_TRACKER_REQUIRED_VALUE_VALIDATION`) when a mandatory field also has "Skip synchronization" on. Core removed that authority in `1cdc9ece91`, **titled "remove unused authority"**, first shipping in 2.42 — so the advice had no effect even before it was removed. The bullet and its trailing paragraph are gone.

**Prose sweep** — 28 remaining references updated:

| File | Edits |
|---|---|
| `configure-programs-in-the-maintenance-app.md` | 17 |
| `configure-metadata-management.md` | 5 |
| `configure-metadata.md` | 4 |
| `about-sharing-of-objects.md` | 2 |

Plus two link fixes: the getting-started table's event and tracker rows merged into one Capture row (replacing a dead `docs.dhis2.org/master/en/user/html/*` URL with a working anchor), and the aggregate Data Entry page's "individual-level data" pointer redirected from Tracker Capture to Capture.

Three of the sweep edits needed judgment rather than find-and-replace:
- The **program rule variable source type** table described genuinely different behaviour per app. Since that distinction survives the app merge, it is now framed as **tracker programs** vs **event programs** rather than collapsed into "Capture app" twice.
- **Field mask / value type hint** notes now read "not in the Capture web app", preserving the point that these are Android-only.
- "such as Android and Tracker Capture" → "such as the Capture and DHIS2 Android Capture apps".

## Deliberately not changed

- The `M_dhis-web-event-capture` / `M_dhis-web-tracker-capture` **module authorities** in `user-authorities.md`. They are stale, but so is most of that table — only 5 of its 36 `M_` rows are still valid — so it needs its own pass.
- `src/stories/*` — historical implementation narratives; rewriting them would falsify the record.
- `src/android/*` — separate app with its own documentation set.
- The legacy **Data Entry** app docs, retired in the same release. Being handled as a separate doc improvement task.

## Verification

- Every anchor defined in the two removed/replaced chapters was checked for inbound references. The only one, `#tracker_capture_app` in `using-the-aggregate-data-entry-app.md`, now points at `#capture_app`.
- `#capture_app` confirmed as the real top-level heading id on the published Capture page (which lives in `dhis2/capture-app`, not this repo), and confirmed to be defined nowhere else, so there is no ambiguity warning.
- The published Capture app docs were checked for stale references to either retired app — there are none.
- Only images that no remaining chapter references were removed.
---

AI Assisted


Philip-Larsen-Donnelly (approved):
LGTM

